PR #5, opened 2015-06-03
Updated POTFILES.in

piotrdrag:
Fixes errors on https://l10n.gnome.org/module/appstream/


ximion:
Yup, thank you for spotting this so quickly and providing a patch!
(merged)

PR #9, opened 2015-10-12
update Catalan translation

rbuj:
I translated appstream to Catalan, I would be very grateful if you could update po/ca.po file with the new translated messages.


ximion:
Okay, done - you might be interested in the Transifex project we have for AppStream, which makes translating and collaborating on translations a bit easier: https://www.transifex.com/freedesktop/appstream/
Thank you for your contribution! :-)

PR #12, opened 2015-10-28
Fixed VAPI compilation.

tintou:
The .gir filename was wrong.


ximion:
Thanks!

tintou:
I've seen that you're the debian maintainer. Would you mind shipping the .vapi file with the -dev package in next releases ?

ximion:
Yes, I can do that (was originally disabled due to a vapigen bug, looks like I forgot to enable it again....)
Expect an upload today :)

ximion:
The Vala API is now available in Debian as well.

PR #15, opened 2015-11-04
Reinitialize the dep11 parser after each file.

tintou:
If more than one file is present in the yaml directory, the parsing will fail after the first file.
Added forgotten reinitialization.


ximion:
Your patch leaks memory, the strings need to be freed first befor you can set them to NULL.

tintou:
Okay I amended the branch as you requested

PR #17, opened 2015-11-15
Component: Added missing documentation

tintou:
Fixed the documentation in several places in AsComponent so that no warning are shown for this object at the documentation generation.


tintou:
Okay, I've added a (skip) to not considerate it in the documentation and in the introspection

PR #18, opened 2015-11-26
doc: Fixed documentation for AsSearchQuery.

tintou:


ximion:
Are you using AsSearchQuery? Because I am thinking of getting rid of that more complex search API, since it seems to be rarely used.

tintou:
I used as_database_find_components to get all the components froma certain category, but your comment made me think that I could simply use as_database_find_components_by_term ("*", category.name)

ximion:
Jup, `as_database_find_components_by_term` will definitely stay...
(Maybe I'll add an easier call for convenience)

PR #19, opened 2016-01-11
lower cmake requirement to 3.1.0

apachelogger:
eaf7229d57ccb77ca29efcfa2bf719ba8fbab568 suggests that it was bumped to
3.2 for CMAKE_CXX_STANDARD, this feature was however introduced in 3.1.

lower the cmake requirement from 3.2 to 3.1 to align with the actual
requirements

tested on 3.1.1


aleixpol:
Makes sense to me, it's better if we can run it on more systems.

PR #22, opened 2016-01-29
Improve Qt5 usage

aleixpol:
Don't rely on QString(const char*)
Prefer splitting with char than with a string.
Leverage Q_GLOBAL_STATIC.


ximion:
Nice, I learned something from this patch :)
The use of QStringLiteral seems to be inconsistent though...

PR #23, opened 2016-02-08
Polishing of the Qt interface

aleixpol:
Don't forget about screenshots.
Allocate before populating for better performance (reduced allocations).


ximion:
That images thing was a dumb bug - we probably need more tests...
Thanks for the patch!

PR #25, opened 2016-02-10
as-cache-builder: Cope with no YAML documents

iainlane:
It's possible that a suite/arch will have no packages with AppStream; we
shouldn't crash in that case.

g_strstr_len will return NULL if the string we're searching for ("---")
is not found - this will be the case if there are only the headers and
no further documents. In that case we should read to the end of the
file.

While doing this, I made this bit (IMO) easier to read by renaming some variables and adding comments, hope you like that.

### DIFF
--- a/src/as-cache-builder.c
+++ b/src/as-cache-builder.c
@@ -182,6 +182,11 @@ as_cache_builder_appstream_data_changed (AsCacheBuilder *builder)
 }
 
 #ifdef APT_SUPPORT
+
+#define YAML_SEPARATOR "---"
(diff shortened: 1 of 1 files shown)


ximion:
Looks good - the empty-DEP-11-doc case shouldn't have been missed. This is pretty much the uglies part of the code, unfortunately just scanning the first part of the doc instead of parsing the YAML correctly is also significantly faster (and we don't want to slow down people's `apt update`).
(For the next release, I plan to make the whole code for importing data from APT a bit smarter anyway).

iainlane:
Thanks!

PR #29, opened 2016-02-19
Check writability

iainlane:
I don't know if this is ready yet - I just gave it light testing, but this should be something like what we discussed over on https://github.com/ximion/appstream/pull/28/

Will be afk mostly over the weekend, so feel free to branch this and push it yourself after doing any needed fixes. :)


ximion:
I implemented this slightly differently, to make the testsuite happy. Also, since AsCacheBuilder is private API, we fortunately can add new parameters to functions as we please :-)

Thanks for the initial patch! Can you maybe test if the current changes completely solve the issue?

Issue #35, opened 2016-05-02
Refresh hangs


neilmayhew:
I have a third-party Ubuntu repo at http://packages.sil.org/ubuntu. I'm using a Xenial test machine. After refreshing sources with `apt-get update`, I run `appstreamcli` as follows:

```
sudo tools/appstreamcli refresh --force --verbose --details
```

It then hangs, using 100% CPU. I've left it as long as 10 minutes and it still doesn't finish. The output is:

```
** (appstreamcli:13992): DEBUG: Refreshing AppStream cache
** (appstreamcli:13992): DEBUG: Reading: /usr/share/app-info/xmls/org.freedesktop.fwupd.xml
** (appstreamcli:13992): DEBUG: Reading: /var/lib/app-info/yaml/archive.ubuntu.com_ubuntu_dists_xenial_main_dep11_Components-amd64.yml.gz
** (appstreamcli:13992): DEBUG: Reading: /var/lib/app-info/yaml/archive.ubuntu.com_ubuntu_dists_xenial-updates_main_dep11_Components-amd64.yml.gz
** (appstreamcli:13992): DEBUG: Reading: /var/lib/app-info/yaml/archive.ubuntu.com_ubuntu_dists_xenial-updates_restricted_dep11_Components-amd64.yml.gz
** (appstreamcli:13992): DEBUG: Reading: /var/lib/app-info/yaml/archive.ubuntu.com_ubuntu_dists_xenial-updates_universe_dep11_Components-amd64.yml.gz
** (appstreamcli:13992): DEBUG: Reading: /var/lib/app-info/yaml/archive.ubuntu.com_ubuntu_dists_xenial-updates_multiverse_dep11_Components-amd64.yml.gz
** (appstreamcli:13992): DEBUG: Reading: /var/lib/app-info/yaml/archive.ubuntu.com_ubuntu_dists_xenial-security_main_dep11_Components-amd64.yml.gz
** (appstreamcli:13992): DEBUG: Reading: /var/lib/app-info/yaml/archive.ubuntu.com_ubuntu_dists_xenial-security_restricted_dep11_Components-amd64.yml.gz
** (appstreamcli:13992): DEBUG: Reading: /var/lib/app-info/yaml/archive.ubuntu.com_ubuntu_dists_xenial-security_universe_dep11_Components-amd64.yml.gz
[...]

PR #41, opened 2016-05-10
Handle errors from the yaml parser

neilmayhew:
This is a more complete treatment of the problem described in #35

---
This change is [Reviewable](https://reviewable.io/reviews/ximion/appstream/41)

### DIFF
--- a/src/as-yamldata.c
+++ b/src/as-yamldata.c
@@ -137,7 +137,7 @@ as_yamldata_free_node (GNode *node, gpointer data)
  * Create GNode tree from DEP-11 YAML document
  */
 static void
-dep11_yaml_process_layer (yaml_parser_t *parser, GNode *data)
+dep11_yaml_process_layer (yaml_parser_t *parser, GNode *data, GError **error)
 {
 	GNode *last_leaf = data;
 	GNode *last_scalar;
@@ -147,7 +147,13 @@ dep11_yaml_process_layer (yaml_parser_t *parser, GNode *data)
 	int storage = YAML_VAR; /* the first element must always be of type VAR */
 
 	while (parse) {
-		yaml_parser_parse (parser, &event);
+		if (!yaml_parser_parse (parser, &event)) {
+			g_set_error (error,
+					AS_METADATA_ERROR,
+					AS_METADATA_ERROR_FAILED,
+					"Invalid DEP-11 file found: YAML parsing error (%s)", parser->problem);
+			break;
+		}
 
 		/* Parse value either as a new leaf in the mapping
 		 * or as a leaf value (one of them, in case it's a sequence) */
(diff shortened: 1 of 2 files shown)


ximion:
This could be the reason why this change wasn't noticed so far: The XML parser ignored garbage after the root node was closed, and for the YAML parser we simply ignored all errors in the format... Bad thing!

src/as-yamldata.c
```diff
@@ -172,7 +178,7 @@ dep11_yaml_process_layer (yaml_parser_t *parser, GNode *data)
 				last_scalar = last_leaf;
 				if (in_sequence)
 					last_leaf = g_node_append (last_leaf, g_node_new (g_strdup ("-")));
-				dep11_yaml_process_layer (parser, last_leaf);
+				dep11_yaml_process_layer (parser, last_leaf, error);
```
ximion: You would need to check if error is NULL after that statement (and then return from the function), because otherwise in the (unlikely) event that the recursion emits multiple errors, we will get a crash when an attempt is made to set an error on a non-NULL GError object.

src/as-yamldata.c
```diff
@@ -1848,7 +1861,7 @@ as_yamldata_parse_distro_data (AsYAMLData *ydt, const gchar *data, GError **erro
 			gboolean header_found = FALSE;
 			GNode *root = g_node_new (g_strdup (""));
 
-			dep11_yaml_process_layer (&parser, root);
+			dep11_yaml_process_layer (&parser, root, error);
```
ximion: Please also check for the error here and return NULL if an error is set (remember to issue a yaml_parser_delete (&parser); before that to not leak memory)

ximion:
I added a few remarks to the patch, you basically would need to exit as early as possible to prevent GError being set twice.

neilmayhew:
I think there are some pre-existing places where memory will be leaked due to the event not being deleted. See the breaks on lines 1861 & 1879 (old numbers).

tests/test-yamldata.c
```diff
@@ -267,14 +278,35 @@ test_yaml_read_languages (void)
...
+					"ID: org.example.Test\n"
+					"\007\n";
+
+	cpt = as_yaml_test_read_data (yamldata_corrupt, &error);
+
+	g_assert_error (error, AS_METADATA_ERROR, AS_METADATA_ERROR_FAILED);
+	g_assert_null (cpt);
+
+	if (error != NULL) {
+		g_assert_true (g_str_has_prefix (error->message, "Invalid DEP-11 file found: YAML parsing error"));
```
neilmayhew: This could be considered too brittle, as it's closely tied to the exact format of the error messages generated.

ximion:
I committed your patch with some modifications to make the GError handling more robust and GLib-ish, and make the parse function return NULL correctly in case of an error.
Also fixed a few GError related minor memory leaks in the testsuite.

Thank you for your patch and general help in debugging this! (Also, it's always awesome to have people provide testsuite testcases with the first patch!)

ximion:
Oh, and btw: I will do a new AppStream release tomorrow, which will include these changes.

neilmayhew:
Thanks for accepting my change. I've not done much programming in the glib environment before, so I'm not surprised it needed some improvements.

It's been a fun and interesting experience working on this, but I hope that everything is fixed now (and SRUs made) so I can get back to my other projects!

ximion:
The SRUs are not in my hands - that's up to Ubuntus stable release maintainers. The AppStream changes are in the queue for proposed-updates, and the appstream-glib changes are pending to be uploaded.
Debian, which is the thing I have more influence on, is ready ;-)

PR #43, opened 2016-05-30
Fix code documentation for as_component_get_id

lucasmoura:
Change documentation on method as_component_get_id from:

Set the unique identifier for this component.

To:

Get the unique identifier for this component.


ximion:
Thanks!

PR #45, opened 2016-06-02
[WIP] Allow AppStream to parse recommendations

lucasmoura:
This MR add the feature to AppStream to allow it to parse recommendation files. A recommendation file is defined by a XML document in the following format:

```xml
<?xml version="1.0"?>
    <recommendations>
        <recommendation>
            <id>vim.desktop</id>
            <because>
                <id>emacs.desktop</id>
            </because>
        </recommendation>
        <recommendation>
            <id>ipython.desktop</id>
            <because>
                <id>eric.desktop</id>
                <id>winpdb.desktop</id>
            </because>
        </recommendation>
    </recommendations>
```

Where *id* is the package being recommended and *because* is the user packages that generated this recommendation.

Currently,  it is already working the construction and parsing of a valid XML recommendation, however there are some necessary improvements to be done:

* There is a lot of code that equals the one on the core AppStream modules and a refactoring is necessary to fix that.
* It is necessary to validate a recommendation object, because if the *recommended* package is not indexed by AppStream, the recommendation object should be invalid.
* Add an appstreamcli function to parse and display the recommendations.
* Validate the XML that represents a recommendation.

Therefore, this MR is manly to collect feedback from the AppStream community regarding this feature.

### DIFF
--- a/src/as-component.c
+++ b/src/as-component.c
@@ -143,6 +143,7 @@ as_component_kind_get_type (void)
 					{AS_COMPONENT_KIND_INPUTMETHOD, "AS_COMPONENT_KIND_INPUTMETHOD", "inputmethod"},
 					{AS_COMPONENT_KIND_ADDON, "AS_COMPONENT_KIND_ADDON", "addon"},
 					{AS_COMPONENT_KIND_FIRMWARE, "AS_COMPONENT_KIND_FIRMWARE", "firmware"},
+					{AS_COMPONENT_KIND_MERGE, "AS_COMPONENT_KIND_MERGE", "merge"},
 					{AS_COMPONENT_KIND_LAST, "AS_COMPONENT_KIND_LAST", "last"},
 					{0, NULL, NULL}
 		};
@@ -178,6 +179,8 @@ as_component_kind_to_string (AsComponentKind kind)
 		return "addon";
 	if (kind == AS_COMPONENT_KIND_FIRMWARE)
 		return "firmware";
+	if (kind == AS_COMPONENT_KIND_MERGE)
+		return "merge";
 	return "unknown";
 }
 
@@ -206,6 +209,8 @@ as_component_kind_from_string (const gchar *kind_str)
 		return AS_COMPONENT_KIND_ADDON;
 	if (g_strcmp0 (kind_str, "firmware") == 0)
 		return AS_COMPONENT_KIND_FIRMWARE;
+	if (g_strcmp0 (kind_str, "merge") == 0)
+		return AS_COMPONENT_KIND_MERGE;
 	return AS_COMPONENT_KIND_UNKNOWN;
 }
 
--- a/src/as-component.c
+++ b/src/as-component.c
@@ -70,6 +70,7 @@ typedef struct
 	GPtrArray		*extensions; /* of string */
 	GPtrArray		*screenshots; /* of AsScreenshot elements */
 	GPtrArray		*releases; /* of AsRelease elements */
+	GPtrArray		*suggestions; /* of AsSuggested elements */
 
 	GHashTable		*provided; /* of int:object */
 	GHashTable		*urls; /* of int:utf8 */
@@ -119,7 +120,8 @@ enum  {
 	AS_COMPONENT_PROJECT_LICENSE,
 	AS_COMPONENT_PROJECT_GROUP,
 	AS_COMPONENT_DEVELOPER_NAME,
-	AS_COMPONENT_SCREENSHOTS
+	AS_COMPONENT_SCREENSHOTS,
+	AS_COMPONENT_SUGGESTIONS
 };
 
 /**
@@ -227,6 +229,7 @@ as_component_init (AsComponent *cpt)
 
 	priv->screenshots = g_ptr_array_new_with_free_func (g_object_unref);
 	priv->releases = g_ptr_array_new_with_free_func (g_object_unref);
+	priv->suggestions = g_ptr_array_new_with_free_func (g_object_unref);
 
 	priv->icons = g_ptr_array_new_with_free_func (g_object_unref);
 	priv->icons_sizetab = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
@@ -268,6 +271,7 @@ as_component_finalize (GObject* object)
 
 	g_ptr_array_unref (priv->screenshots);
 	g_ptr_array_unref (priv->releases);
+	g_ptr_array_unref (priv->suggestions);
 	g_hash_table_unref (priv->provided);
 	g_hash_table_unref (priv->urls);
 	g_hash_table_unref (priv->languages);
@@ -399,6 +403,22 @@ as_component_add_release (AsComponent *cpt, AsRelease* release)
 
 	releases = as_component_get_releases (cpt);
 	g_ptr_array_add (releases, g_object_ref (release));
+}
+
+/**
+ * as_component_add_suggestion:
+ * @cpt: a #AsComponent instance.
+ * @suggested: The #AsSuggested to add
+ *
+ * Add an #AsSuggested to this component.
+ **/
+void
+as_component_add_suggestion (AsComponent *cpt, AsSuggested* suggested)
+{
+	GPtrArray* suggestions;
+
+	suggestions = as_component_get_suggestions (cpt);
+	g_ptr_array_add (suggestions, g_object_ref (suggested));
 }
 
 /**
@@ -1361,6 +1381,22 @@ as_component_get_compulsory_for_desktops (AsComponent *cpt)
 }
 
 /**
+ * as_component_get_suggestions:
+ * @cpt: a #AsComponent instance.
+ *
+ * Get a list of associated suggestions.
+ *
+ * Returns: (element-type AsSuggested) (transfer none): an array of #AsSuggested instances
+ */
+GPtrArray*
+as_component_get_suggestions (AsComponent *cpt)
+{
+	AsComponentPrivate *priv = GET_PRIVATE (cpt);
+
+	return priv->suggestions;
+}
+
+/**
  * as_component_set_compulsory_for_desktops:
  * @cpt: a #AsComponent instance.
  * @value: (array zero-terminated=1): the array of desktop ids.
@@ -2234,6 +2270,9 @@ as_component_get_property (GObject * object, guint property_id, GValue * value, 
 		case AS_COMPONENT_SCREENSHOTS:
 			g_value_set_boxed (value, as_component_get_screenshots (cpt));
 			break;
+		case AS_COMPONENT_SUGGESTIONS:
+			g_value_set_boxed (value, as_component_get_suggestions (cpt));
+			break;
 		default:
 			G_OBJECT_WARN_INVALID_PROPERTY_ID (object, property_id, pspec);
 			break;
@@ -2411,6 +2450,14 @@ as_component_class_init (AsComponentClass * klass)
 	g_object_class_install_property (object_class,
 					AS_COMPONENT_SCREENSHOTS,
 					g_param_spec_boxed ("screenshots", "screenshots", "screenshots", G_TYPE_PTR_ARRAY, G_PARAM_STATIC_NAME | G_PARAM_STATIC_NICK | G_PARAM_STATIC_BLURB | G_PARAM_READABLE));
+	/**
+	 * AsComponent:suggestions: (type GPtrArray(AsSuggested)):
+	 *
+	 * An array of #AsSuggested instances
+	 */
+	g_object_class_install_property (object_class,
+					AS_COMPONENT_SUGGESTIONS,
+					g_param_spec_boxed ("suggestions", "suggestions", "suggestions", G_TYPE_PTR_ARRAY, G_PARAM_STATIC_NAME | G_PARAM_STATIC_NICK | G_PARAM_STATIC_BLURB | G_PARAM_READABLE));
 }
 
 /**
--- a/src/as-suggested.c
+++ b/src/as-suggested.c
@@ -1,0 +1,214 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2016 Lucas Moura <lucas.moura128@gmail.com>
+ *
+ * Licensed under the GNU Lesser General Public License Version 2.1
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 2.1 of the license, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * SECTION: as-suggested
+ * @short_description: Descibe which componentes were responsible for
+ *					   a given package recommendation.
+ * @include: appstream.h
+ */
+
+#include "config.h"
+
+#include "as-suggested.h"
+
+typedef struct
+{
+	AsSuggestedKind	kind;
+	GPtrArray		*cpts_id; /* of string */
+} AsSuggestedPrivate;
+
+G_DEFINE_TYPE_WITH_PRIVATE (AsSuggested, as_suggested, G_TYPE_OBJECT)
+#define GET_PRIVATE(o) (as_suggested_get_instance_private (o))
+
+/**
+ * as_suggested_kind_to_string:
+ * @kind: the %AsSuggestedKind.
+ *
+ * Converts the enumerated value to an text representation.
+ *
+ * Returns: string version of @kind
+ **/
+const gchar*
+as_suggested_kind_to_string (AsSuggestedKind kind)
+{
+	if (kind == AS_SUGGESTED_KIND_HEURISTIC)
+		return "heuristic";
+	if (kind == AS_SUGGESTED_KIND_UPSTREAM)
+		return "upstream";
+
+	return "unknown";
+}
+
+/**
+ * as_suggested_kind_from_string:
+ * @kind_str: the string.
+ *
+ * Converts the text representation to an enumerated value.
+ *
+ * Returns: a #AsSuggestedKind or %AS_SUGGESTED_KIND_UNKNOWN for unknown
+ **/
+AsSuggestedKind
+as_suggested_kind_from_string (const gchar *kind_str)
+{
+	if (g_strcmp0 (kind_str, "heuristic") == 0)
+		return AS_SUGGESTED_KIND_HEURISTIC;
+	if (g_strcmp0 (kind_str, "upstream") == 0)
+		return AS_SUGGESTED_KIND_UPSTREAM;
+
+	return AS_SUGGESTED_KIND_UNKNOWN;
+}
+
+/**
+ * as_suggested_finalize:
+ **/
+static void
+as_suggested_finalize (GObject *object)
+{
+	AsSuggested *suggested = AS_SUGGESTED (object);
+	AsSuggestedPrivate *priv = GET_PRIVATE (suggested);
+
+	g_ptr_array_unref (priv->cpts_id);
+
+	G_OBJECT_CLASS (as_suggested_parent_class)->finalize (object);
+}
+
+/**
+ * as_suggested_init:
+ **/
+static void
+as_suggested_init (AsSuggested *suggested)
+{
+	AsSuggestedPrivate *priv = GET_PRIVATE (suggested);
+
+	priv->cpts_id = g_ptr_array_new_with_free_func ((GDestroyNotify) g_object_unref);
+}
+
+/**
+ * as_suggested_new:
+ *
+ * Creates a new #AsSuggested.
+ *
+ * Returns: (transfer full): a new #AsSuggested
+ **/
+AsSuggested*
+as_suggested_new (void)
+{
+	AsSuggested *suggested;
+	suggested = g_object_new (AS_TYPE_SUGGESTED, NULL);
+	return AS_SUGGESTED (suggested);
+}
+
+/**
+ * as_suggested_class_init:
+ **/
+static void
+as_suggested_class_init (AsSuggestedClass *klass)
+{
+	GObjectClass *object_class = G_OBJECT_CLASS (klass);
+	object_class->finalize = as_suggested_finalize;
+}
+
+/**
+ * as_suggested_get_kind:
+ * @suggested: a #AsSuggested instance.
+ *
+ * Gets the suggested kind.
+ *
+ * Returns: the #AssuggestedKind
+ **/
+AsSuggestedKind
+as_suggested_get_kind (AsSuggested *suggested)
+{
+	AsSuggestedPrivate *priv = GET_PRIVATE (suggested);
+	return priv->kind;
+}
+
+/**
+ * as_suggested_set_kind:
+ * @suggested: a #AsSuggested instance.
+ * @kind: the #AsSuggestedKind, e.g. %AS_SUGGESTED_KIND_HEURISTIC.
+ *
+ * Sets the suggested kind.
+ **/
+void
+as_suggested_set_kind (AsSuggested *suggested, AsSuggestedKind kind)
+{
+	AsSuggestedPrivate *priv = GET_PRIVATE (suggested);
+	priv->kind = kind;
+}
+
+/**
+ * as_suggested_get_components_id:
+ * @suggested: a #AsSuggested instance.
+ *
+ * Get a list of components id that generated the suggestion
+ *
+ * Returns: an array of components id
+ */
+GPtrArray*
+as_suggested_get_components_id (AsSuggested* suggested)
+{
+	AsSuggestedPrivate *priv = GET_PRIVATE (suggested);
+
+	return priv->cpts_id;
+}
+
+
+/**
+ * as_suggested_add_component_id:
+ * @suggested: a #AsSuggested instance.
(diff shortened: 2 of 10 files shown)


ximion:
Nice! I'll be a bit busy the next days (KDE AppStore meeting and other things), but I think I can give it a quick review tonight. One obvious thing to resolve would be fixing the test issues ;-) (minor stuff, as far as I can see).
Some quick initial questions:

1) Where is that recommendation data supposed to come from? Are upstream projects supposed to ship it in their metainfo file? Or doe distributions seed that stuff?

2) Regarding the XML example:
```XML
<recommendations>
        <recommendation>
            <recommended>vim.desktop</recommended>
            <because>
                <pkg>emacs.desktop</pkg>
            </because>
```
You want to define a condition here, right? So, something like "if Vim present, suggest Emacs". The "because" should therefore probably be an "if". Also, I think the whole block is a bit strange... Why not put the recommendation into the component-block of the thing recommending it? So, something like:
```XML
<component type="desktop">
  <id>emacs.desktop</id>
  <recommendations>
    <recommendation>emacs.desktop</recommendation>
  </recommendations>
</component>
```
That would look far more simple to me, and would prevent the need for declaring logic in XML. Aside from that, your `pkg` tag is actually referring to an AppStream-ID (probably a leftover from when it was package-based).

I need to read the code this evening to fully understand the concept behind this and give some better feedback though ;-)

lucasmoura:
No problem. I will look why the build is failing. It is currently presenting this error:

gpg: no valid OpenPGP data found.var/lib/app-inf

But I will look it up to fix it.

1) Originally, I thought that the recommendation application could generate the file and place it a common place for AppStream to parse the recommendations. Something like /usr/share/recommendations. But I still don't know if that is the best approach.

2) I think I have not explained the *because* tag properly. The recommendation would not be something like *if the user has emacs.desktop then recommends vim.desktop*, it would be *because the user already has emacs.desktop then recommends vim.desktop*. The recommender application would be the one responsible for asserting which packages the user has and control this representation.

But you are right, the tag names provided by your XML file are way better than the ones I have provided. I will create an XML file with better names and update the code accordingly.

ximion:
Sidenote: I *really* want a reply function in the GitHub UI! Making replys manually is annoying...

> gpg: no valid OpenPGP data found.var/lib/app-inf
> But I will look it up to fix it.

No worries, that was a temporary failure due to the LLVM APT repo not being available. I took that as an argument to make the whole build process happen in a Docker container now, which allows us to be independent of the outdated Travis base system.
So, to fix this for you, just rebase on master.

> The recommender application would be the one responsible for asserting which packages the user has and control this representation.

Okay, then why have this because information in the first place? Wouldn't simply dropping a list of recommended AppStream-IDs into some directory in `/var/cache` be more useful? (and have AS pick that up then)

lucasmoura:
Okay, I have applied the rebase and I will check again if there is another problem during the build process.

But about the *because* tag. Originally, I though that it would be good for an user to know which of his packages generated the one being recommended. It would made the recommendation more trustworthy, since the user would see the reason behind it. Therefore, I thought that the applications that would use this AppStream data could display this extra information for their users.

However, maybe on the context of AppStream this extra information is really not necessary or I can make it optional.

lucasmoura:
I have added some tests for invalid recommendation XML files

ximion:
Can you please give me a complete exmplanation why you need all of that new XML stuff?
If the AppRecommender runs as independent tool, why can't it just throw out a list of AppStream IDs that libappstream can then read and flag specific components as "recommended"?
All the things you write seem to be a bit overcomplicated for this task, TBH.

lucasmoura:
Yes, sure. The reason I am creating the *because* tag is an attempt to allow users to better understand why a recommendation is being done. For example, suppose that I just use a list of AppStream *ids* as a recommendation file. If any application wants to use AppStream to show these data, it will only show the recommended packages for the user, but it will not be possible to show why that package was recommended in the first place.

Therefore, I believe that using the *because* tag can make the recommendation more trustworthy for the user.

But if you think that tag should not be treated on AppStream or it will not be as useful, I can remove it with no problem :)

aleixpol:
I think that what @ximion  means is that the format itself could be reduced somewhat. In the end, we're just offering a pair of packages such as. It's about offering pairs such as `gdb recommends gcc` and `vim+x11 recommends kate`. Finding ways to easily offer those could help adoption.

FWIW, do recommendations be directional? Does `emacs recommends vim` inherently mean the opposite?

lucasmoura:
@aleixpol True, that makes sense. I will try to think of a better representation for a recommendation file. 

About the second question, I don't know for sure if the recommendations are directional. Since the user profile is a combination of the most significant terms of the user manual installed packages, I cannot answer for sure if swapping a recommended package for an installed one, would then make the recommender system to recommended the installed package. But again, I may be wrong on that, since I have never looked at this situation closely.

@ximion I think I have misunderstood your comments, I believe you are not talking about the *because* tag, but the new code I am adding to the feature. Basically,  the only new feature I have added is the one that checks if a recommendation object actually holds a package indexed by AppStream. The other code I have written is just to make the test suite more robust.

ximion:
@lucasmoura There are basically two fundamental things I am criticising:
 * The code changes you do are massive and duplicate most of the XML and AppStream parsing logic, while we already do have logic to parse AppStream-style XML data and enhance existing data with it.
 For example, if you have an AppStream component with ID "org.example.foobar.desktop" defined, and then have the generator read XML like:

 ```XML
<component type="desktop">
  <id>org.example.foobar.desktop</id>
  <recommendation>
    <because>XYZ</because>
  </recommendation>
</component>
```
 The XML reader you yet a recommended=True flag on the original component type, and apps could show that information immediately. It would also greatly reduce code duplication.
 Right now, you create an `AsRecommendation` class that is separate, and I am not sure if that is actually needed. I am also not sure if we need information on why something was recommended at all, because users in software-centers don't really care about that. In case someone cares, they could ask apprecommender for that intel directly.

 * I am not yet sure about the whole because-of stuff and the general design. It would be nice if you could outline who is generating what and what exactly you have in mind how this thing should work. I can guess from your code, but that's really not enough ;-) By doing so, you'd also safe yourself from doing work which doesn't get merged - I really hate to see people putting lots of working into new code and then not being able to merge a PR because it doesn't fit the design of the application.
 So, what I would like to see is something like "AppRecommender reads list of packages, resolves them to Components via libappstream, produces a list of recommended Component-IDs based on that data and writes a list of them and why they were recommended to `/var/cache/app-info/recommendations` (or to some path in $HOME). AppStream itself will read that file, mark recommended applications/components as such and allow software-center GUIs to display that information. AppStream should allow SC apps to display why something was recommended by parsing the machine-readable AppRecommender XML and presenting it to the SCs."

**Comment by lucasmoura** [comment]
@ximon Now I have completely understood you point. Also, I am really sorry for taking that long to understand it. Yes, using the component XML infrastructure to make a recommendation would definitely be better. I was actually studying ways to solve the problem of code duplicate, but I think that approach solves that problem nicely. Therefore, I will change my code for this new format.

For the second part. Yes, I will make a step by step guide of how this data will be generated and consumed. I strongly believe that I will make that guide tonight.

But thanks again for the patience and sorry one more time for my delay on understanding your point.

**Comment by lucasmoura** [comment]
@ximion This is the step by step guide on how I believe the whole recommendation feature would work.

1) During the package installation of AppRecommender will create an user profile based on the user manual installed packages.For the context of software centers, the manual installed packages will be filtered to allow only AppStream componentes to be used.

2) After that, AppRecommender will perform a recommendation and generate a XML file containing the component being recommended and the user components that generated that recommendation.

3) The XML file would be similar as the component XML file, however, it will only possess an *id* and *recommendations* tag. The *id* tag would hold the actual package being recommended and the *recommendations* tag would be used to store the packages that generated the *id* recommendation. AppRecommender will be responsible to provide both informations.

4) AppRecommender will then place this XML file on /usr/share/app-info/recommendations

5) After that, this file will be updated via a crontab, which will run an AppRecommender command to update that XML file. This command would run weekly.

6) Once the file is in the proper location, AppStream will be able to read the file and loads all the recommended components on memory.
* (I still don't know if it is better for AppStream to always read this file and create the objects dinamically, or if I should store them in a Xapian database)

7) If there are more than one file it the /usr/share/app-info/recommendations, AppStream would read the most recent file.

8) AppStream method would them be used to parse the recommendations and loads them on memory as components. It would be similar to the method *as_database_get_all_components*.

9) A Software Center to use this feature would them use this feature to load the recommended components and display then to the users.

10) The recommendation objects would contain the package being recommended and the packages that generated that recommendation.

-----
That is basically my general idea on how that whole integration will work. However, I believe that if the *because* tag will be present on the XML file, the _AsRecommendation_ class would be needed, I could use the *pkgnames* variable on _AsComponent_ to hold that information, but I don't think it is a good idea. But since you stated that Software Center users will not care for that additional information at all, maybe it is best to drop it.

**Comment by lucasmoura** [comment]
@ximion I have  implemented a new solution using now AppStream already implemented methods to parse the recommendation file.

The feature works as specified in my the guide I have provided. Basically, now I am using the as-data-pool to parse the recommendation files. Furthermore, the *AsRecommendation* object now only serve the purpose of retrieving the recommendation components and nothing else.

I have also removed the meta information about why a recommendation was generated and just used the list of component ids as a recommendation file.

I still don't know if the verification if the recommendation file holds a valid component should be performed on the *AsRecommendation* object, or if it should be done by the user who wants to use that feature.

**Comment by ximion** [comment]
To:

1) Sounds good to me - although it would probably be useful to operate directly on AppStream-IDs in $future, because these include software from bundling sources as well.

2) What concept are you referring to with "user components"?

3) A component-type "recommendation" doesn't make sense, because a component type is *always* a specific type of installable software. Recommendation is a property of a component. If we decide to drop the why-is-it-recommended part, Apprecommender could also simply output a plaintext file listing AppStream-IDs - those could be read faster. We would then set a `recommended?` boolean in `AsComponent` for recommended apps.

4 and 5) Sounds good

6) AppStream will read the XML and use Xapian as cache - that's the basic concept. I will also only update the cache if something has changed.

7) I don't like that "most recent file" logic - can't there just be one file? Why do you expect multiple files to be there? Also, the default mode of action would be to read all files use the information there. If a file is no longer relevant, it should be deleted.

8) Recommendation component-type: Nonononono, see 3) - a recommendation is a *property* of an `AsComponent`.

9) Jup

10) You mean the component-id being recommended, not the package, right?

---
Independent from that: Will you be at Debconf? If so, we can discuss this there and work on it.

**Comment by ximion** [comment]
Calling the software-center developers: @hughsie @aleixpol @tintou @ikeydoherty
Please don't read the whole bug report, I am highlighting you for this very simple question:
Would you in your software-center, when you display "featured" or "recommended" apps (with the recommendations based on what else the user has installed) also display *why* these apps were recommended? (As in "Vim is recommended because you installed gcc, devhelp and cmake")

I would say this information is nothing which the user cares much about, because it's nothing which is actionable and is purely for the people who are curious about how the recommendations work. If they get "bad" recommendations, that info won't help, if they get good recommendations, they won't care. Also, the info can still be retrieved via CLI tools, and might clutter software center UIs, which already have to show quite a lot of information in an appealing way.

So, what do you think? I am very reluctant to let the complex "X because Y and not Z" logic into the spec and have it implemented in libappstream. With my maintainer's hat on, I think this will bite me later, so I would not want this feature in AppStream itself. Of course, my opinion on it will change if you say you want to expose this feature to end users and would like to have it in SC UIs.

**Comment by lucasmoura** [comment]
@ximion 

Answering a couple of questions:

2) The "user components" would be the user packages that generated the recommendation and that are also indexed on AppStream.

3) Fair enough. I created the *recommendation* type to allow the *as_component_is_valid* to work properly, but now I see that it doesn't make sense in the context of a component. Using the *recommended?* variable on a component would work way better.

6) Okay

7) Maybe I have overthought that step. I was thinking if the user has more than one recommender application that would like to integrate with AppStream. But allowing a single file would be a better approach.

10) Yes, my mistake there.

----
I will be attending DebConf this year, and yes, it will be great if we could discuss and work on this there :)

**Comment by colomar** [comment]
@ximion Answering for Discover because I'm the one responsible for its user interface: While I would not display the "why" by default (to reduce clutter), I would want to make it easily accessible on demand. The reason being that users may gauge the usefulness of a recommendation based on its source.
If an application is recommended to me based on one which I use regularly, I may be more likely to heed it compared to one I got based on some app I installed just to try it out or for some specific task and have rarely - if ever - used it since (assuming people do not uninstall applications they do not use)

**Comment by hughsie** [comment]
So, I see this useful in the same way I see extends being used, e.g.

    <component>
       <id>gimp.desktop</id>
      <recommend>mypaint.desktop</recommend>
      <recommend>krita.desktop</recommend>
    </component>

I have no idea why you'd want to encode the "why" in the data whatsoever. If we do, something like:

   <recommend reason="same-developer">colorhug-client.desktop</recommend>

...could work i guess, and would allow us to do things like the play store.

**Comment by danrabbit** [comment]
Hm I'm not sure why you'd want to hate code this into the spec. It seems that would mean apps developers have to be aware of every other possible package.

But regarding why that info is useful, as says above the user might gauge the usefulness of the recommendation based on their feelings of another app. But also, this information is valuable if a feature exists for the user to mark an app to not be used for recommendations. Think Amazon. I'm getting recommendations that I think are not relevant and I want to explain to the computer why I think they're not relevant.

**Comment by ximion** [comment]
Notes from Debconf:
Allow upstream metainfo files to include recommendations in form of
```XML
<component type="desktop">
  <id>org.example.Foo.desktop</id>
  <name/>
  <summary/>
  <description/>
  <suggests type="upstream">
    <id>org.example.Bar.desktop</id>
    <id>...</id>
  </suggests>
</component>
```
The type might be omitted, and metainfo files are only allowed to include type "upstream" suggestions.

Allow distro metadata files to enhance existing components, add a "heuristic" type for AppRecommender:

```XML
<components>
  <component type="partial">
    <id>org.example.Foo.desktop>
    <suggests type="heuristic">
      <id>org.example.XYZ</id>
      ...
    </suggests>
  </component>
  ...
</components>
```

This way, software centers have basic informtion on where a recommendation is coming from, and also AppRecommender doesn't need to invent its own format and can just use standard AppStream metadata (as soon as this spec extension is standardized).

lucasmoura:
If the package being recommended is not the one receiving the *suggests* tag,  I think that there might be some problems. A recommendation in AppRecommender is generated by the combination of multiple packages. Therefore a single package recommending others may not be a good representation of the process.

Maybe the *suggests* tag can turn into something like a *suggested-by* tag and if a component has this tag populated it can be assumed that it is a recommendation. But I don't know if that would the best  approach here.

hughsie:
I think it's fine the AppStream "app A suggests app B" is encoded into the file. I don't think we want to put the internal data structure or exact reason in the output data.

lucasmoura:
Fair enough. I am only a little bit uncertain about making AppRecommender produce a result such as "app A recommends app B", because in reality would be something like "app A, B, C and D recommends E".  I am afraid that by doing so, considering only one package, the recommendation produced by AppRecommender may be a little misleading.

But I can check that out and see if it is true.

lucasmoura:
@ximion I thing I have created the necessary changes to allow AppStream to handle suggestions metadata. However, I am not sure if I am testing the best way I can or if the tests are in the right file. If there is still anything wrong with my solution, please let me know :)

lucasmoura:
I just forgot to say that, if everything is alright with this solution, I can start working on its documentation as well.

lucasmoura:
I have added the "merge" type for a component class. I have also made a verification on the
**as_data_pool_merge_components** method, to only merge components if the **src_cpt** has the **merge** type.

src/as-suggested.c
```diff
@@ -0,0 +1,214 @@
...
+ */
+
+#include "config.h"
+
+#include "as-suggested.h"
+
+typedef struct
+{
+	AsSuggestedKind	kind;
+	GPtrArray		*cpts_id; /* of string */
```
ximion: Should rather be `cpt_ids` of utf8

src/as-suggested.c
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * as_suggested_kind_from_string:
+ * @kind_str: the string.
+ *
+ * Converts the text representation to an enumerated value.
+ *
+ * Returns: a #AsSuggestedKind or %AS_SUGGESTED_KIND_UNKNOWN for unknown
+ **/
+AsSuggestedKind
+as_suggested_kind_from_string (const gchar *kind_str)
```
ximion: For a not-set string (NULL) we should assume UPSTREAM.

src/as-suggested.c
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * as_suggested_init:
+ **/
+static void
+as_suggested_init (AsSuggested *suggested)
+{
+	AsSuggestedPrivate *priv = GET_PRIVATE (suggested);
+
+	priv->cpts_id = g_ptr_array_new_with_free_func ((GDestroyNotify) g_object_unref);
```
ximion: This will explode, you don't want `g_object_unref` but `g_free`.

src/as-suggested.c
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * as_suggested_get_components_id:
+ * @suggested: a #AsSuggested instance.
+ *
+ * Get a list of components id that generated the suggestion
+ *
+ * Returns: an array of components id
+ */
+GPtrArray*
+as_suggested_get_components_id (AsSuggested* suggested)
```
ximion: get_component_ids - it's multiple... (otherwise I would expect some kind of filter to just return one)

src/as-suggested.c
```diff
@@ -0,0 +1,214 @@
...
+ * @suggested: a #AsSuggested instance.
+ * @cpt_id: The component id to add
+ *
+ * Add a component id to this suggested object.
+ **/
+void
+as_suggested_add_component_id (AsSuggested *suggested, gchar* cpt_id)
+{
+	AsSuggestedPrivate *priv = GET_PRIVATE (suggested);
+	g_ptr_array_add (priv->cpts_id, cpt_id);
```
ximion: Uuuh, you need to strdup here, otherwise ownership of the pointer is transferred to the ptr-array, which will free it later. This isn't good.

src/as-suggested.c
ximion: On second thought, I also wonder if a GHashTable isn't a better idea here, since that will prevent us from having duplicate entries...

src/as-suggested.c
```diff
@@ -0,0 +1,214 @@
...
+ * as_suggested_is_valid:
+ * @suggested: a #AsSuggested instance.
+ *
+ * Check if the essential properties of this suggestion are
+ * populated with useful data.
+ *
+ * Returns: TRUE if the suggestion data was validated successfully.
+ */
+gboolean
+as_suggested_is_valid (AsSuggested *suggested)
```
ximion: A method like this isn't necessary - but it also doesn't hurt, so let's have it :)

src/as-suggested.c
```diff
@@ -0,0 +1,214 @@
...
+	AsSuggestedKind stype;
+
+	AsSuggestedPrivate *priv = GET_PRIVATE (suggested);
+
+	stype = priv->kind;
+	if (stype == AS_SUGGESTED_KIND_UNKNOWN)
+		return FALSE;
+
+	if (priv->cpts_id->len != 0)
+		ret = TRUE;
```
ximion: Why so complicated with a variable? Just return TRUE and otherwise FALSE.

src/as-suggested.c
```diff
@@ -0,0 +1,214 @@
...
+	priv->kind = kind;
+}
+
+/**
+ * as_suggested_get_components_id:
+ * @suggested: a #AsSuggested instance.
+ *
+ * Get a list of components id that generated the suggestion
+ *
+ * Returns: an array of components id
```
ximion: This part is missing GIR annotations.

src/as-suggested.h
```diff
@@ -0,0 +1,79 @@
...
+const gchar			*as_suggested_kind_to_string (AsSuggestedKind kind);
+
+AsSuggested			*as_suggested_new (void);
+
+AsSuggestedKind     as_suggested_get_kind (AsSuggested *suggested);
+void				as_suggested_set_kind (AsSuggested *suggested,
+						AsSuggestedKind kind);
+
+GPtrArray*			as_suggested_get_components_id (AsSuggested *suggested);
+void				as_suggested_add_component_id (AsSuggested *suggested, gchar* cpt_id);
```
ximion: Should be a `const gchar*` as parameter.

src/as-component.c
```diff
@@ -402,6 +406,22 @@ as_component_add_release (AsComponent *cpt, AsRelease* release)
...
 
 /**
+ * as_component_add_suggestion:
+ * @cpt: a #AsComponent instance.
+ * @suggested: The #AsSuggested to add
+ *
+ * Add an #AsSuggested to this component.
+ **/
+void
+as_component_add_suggestion (AsComponent *cpt, AsSuggested* suggested)
```
ximion: Should likely be add_suggested for consistency.

src/as-component.c
```diff
@@ -402,6 +406,22 @@ as_component_add_release (AsComponent *cpt, AsRelease* release)
...
+{
+	GPtrArray* suggestions;
+
+	suggestions = as_component_get_suggestions (cpt);
```
ximion: No need for this, just access the variable directly.

ximion:
I fixed up the first two patches and committed the result to master.
Can you please rebase the remaining bits on top of latest master? (there were quite a few changes, for example the inclusion of the merge component, and a clean set of patches will be easier to review).

Bonus points for adding some validator hints (e.g. having a `type=heuristic` component in metainfo files is an error).

src/as-xmldata.c
```diff
@@ -443,6 +443,50 @@ as_xmldata_process_screenshots_tag (AsXMLData *xdt, xmlNode* node, AsComponent*
...
+	AsSuggested *suggested = NULL;
+	AsSuggestedKind suggested_kind;
+	gchar *node_name;
+	gchar *type_str;
+	gchar *content;
+
+	g_return_if_fail (xdt != NULL);
+	g_return_if_fail (cpt != NULL);
+
+	suggested = as_suggested_new ();
```
ximion: You are leaking memory here.

src/as-xmldata.c
```diff
@@ -443,6 +443,50 @@ as_xmldata_process_screenshots_tag (AsXMLData *xdt, xmlNode* node, AsComponent*
...
+	type_str = (gchar*) xmlGetProp (node, (xmlChar*) "type");
```
ximion: ...and here (needs to be free'd)

src/as-xmldata.c
```diff
@@ -443,6 +443,50 @@ as_xmldata_process_screenshots_tag (AsXMLData *xdt, xmlNode* node, AsComponent*
...
+
+	for (iter = node->children; iter != NULL; iter = iter->next) {
+		/* discard spaces */
+		if (iter->type != XML_ELEMENT_NODE)
+			continue;
+
+		node_name = (gchar*) iter->name;
+
+		if (g_strcmp0 (node_name, "id") == 0) {
+			content = as_xmldata_get_node_value (xdt, iter);
```
ximion: As well as here (actually, this is probably the biggest leak, since this needs to be freed on each iteration)

src/as-xmldata.c
```diff
@@ -443,6 +443,50 @@ as_xmldata_process_screenshots_tag (AsXMLData *xdt, xmlNode* node, AsComponent*
...
+
+			if (content != NULL)
+				as_suggested_add_component_id (suggested, content);
+		}
+
+		if (as_suggested_is_valid (suggested))
+			as_component_add_suggestion (cpt, suggested);
```
ximion: You are adding the `AsSuggested` again and again here, once per ID...

tests/test-pool.c
```diff
@@ -193,6 +193,70 @@ test_pool_read ()
 	g_assert_cmpuint (as_release_get_size (rel, AS_SIZE_KIND_DOWNLOAD), ==, 0);
 }
 
+void
+test_merge_components ()
+{
+	guint i;
+
+	g_autoptr(GPtrArray) cpts = NULL;
```
ximion: Unused variable - weird that the CI didn't complain about this...

tests/samples/distro/xmls/suggestions.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<components version="0.8" origin="jessie">
+  <component type="desktop">
+    <id>test.desktop</id>
```
ximion: Merge components need some component to merge their data into - otherwise they don't exist in the pool.
Since there is no test.desktop, this would fail.

tests/test-pool.c
```diff
@@ -193,6 +193,70 @@ test_pool_read ()
...
+		g_assert_cmpstr (cpt_id, ==, expected_cpts_id[i]);
+	}
+
+	cpt = as_data_pool_get_component_by_id (dpool, "test1.desktop");
+	suggestions = as_component_get_suggestions (cpt);
+	suggested = (AsSuggested *) g_ptr_array_index (suggestions, 0);
+	cpts_id = as_suggested_get_components_id (suggested);
+	suggested_kind = as_suggested_get_kind (suggested);
+
+	expected_cpts_id = g_new0 (gchar*, 3);
```
ximion: Memory leak... You need to free the strv first.

ximion:
The contents of this PR are now merged - there is still work to do, mainly support for writing the suggests tag and for the YAML part is completely missing.
Also, there are no tests for reading the data from an XML file / writing it yet.

ximion:
Thanks for submitting this PR!

lucasmoura:
Sorry for that huge amount of memory leaks. I will be more careful on my next merge request. I think I will now allow AppRecommender to write the XML files and look into the YAML part as well.

Thanks for the patience on reviewing this feature :)

ximion:
I already took care of the YAML world (and pretty much all other bits) - the remaining piece is getting the merge logic work for (almost) all fields, see the `as_merge_components` function in `AsDataPool`.

I hope I can do a new AppStream release ASAP.
Please pay attention that @hughsie and I changed how component merges are done after a brief discussion on IRC, check the example file in the testsuite for how it's done now :)

The libappstream library is capable of writing the necessary XML or YAML files for merge components now.

PR #56, opened 2016-07-27
Different approach for storing the icon size and url relation

aleixpol:
This way we can return it as a block so that the client can actually
introspect which icons are available.

The code I'm referring to looks like this:
```
{
    QIcon ret;

    const auto icons = m_appdata.iconUrls();
    if (icons.isEmpty())
        return m_appdata.name();
    else {
        for (auto it = icons.constBegin(), itEnd = icons.constEnd(); it!=itEnd; ++it) {
            if (it->isLocalFile())
                ret.addFile(it->toLocalFile(), it.key());
        }
    }
    return ret;
}
```

### DIFF
--- a/qt/component.cpp
+++ b/qt/component.cpp
@@ -22,9 +22,15 @@
 #include <QSharedData>
 #include <QStringList>
 #include <QUrl>
+#include <QMap>
 #include <QMultiHash>
 
 using namespace Appstream;
+
+static bool operator<(QSize a, QSize b)
+{
+    return a.width() < b.width() || (a.width() == b.width() && a.height() < b.height());
+}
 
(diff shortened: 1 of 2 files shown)


ximion:
Looks good to me, but I wonder if this solution is ideal for the future... Libas has an own `AsIcon` class, but mainly because it doesn't have a `QUrl` (because plain C and GLib).
Anyway, it's definitely a useful improvement for the present!

aleixpol:
TBH, what worries me the most at the moment is the ::isLocalFile, will have to look into how to do it when it's not local. Still not a problem as of today.

Thanks!

ximion:
The thing I thought about the most was stock icons, which are no URLs at all.... :D

PR #64, opened 2016-08-17
as_flags_value is broken

iainlane:
Not sure what you intended here - but the test case doesn't pass, and it doesn't work in asgen either. The values get shifted twice, which means the test never works.

This fixes it for me, but if you had something else in mind then please do that instead. The second commit is just a typo fix.


ximion:
Hmm, the tests definitely didn't fail (CI approves), but this is an error regardless.
I copied those lines from PackageKit, and PKs PkBitfields are special in some regards.

(Fixed now in master, by dropping the _value method)

PR #81, opened 2016-10-03
Replace refresh-index with refresh in help text

larryprice:
I noticed the "--help" text was no longer accurate in recent versions of appstream. I replaced the "refresh-index" text with the "refresh" command. This could also or alternatively show "refresh-cache" if preferred, but it seems the pattern is to show the basic command in the "help" text and alternatives in the man page.


ximion:
Hmm, this should actually really be `refresh-cache` there - we show the less generic command, because that won't change in future (`refresh` might get additional meanings in future, e.g. "download fresh metadata from the web").
Can you adjust your patch?

PR #82, opened 2016-10-04
Fix database dtor crash

apachelogger:
without this an unopened database goes boom because the pointer points to garbage


aleixpol:
+1 as soon as the tests pass. @ximion any idea what happened there?

apachelogger:
thar she blows. build fixed. all lovely again. I am calling it a day :P

PR #83, opened 2016-10-25
Directly offer a QIcon from the component

aleixpol:
Instead of giving the different pieces (which could also be accessible
through the API)

### DIFF
--- a/qt/CMakeLists.txt
+++ b/qt/CMakeLists.txt
@@ -5,7 +5,7 @@ include(CMakePackageConfigHelpers)
 include(GenerateExportHeader)
 add_compiler_export_flags()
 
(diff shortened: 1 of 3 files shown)


ximion:
This introduces a dependency on GUI stuff which I was very careful to avoid so far...
I guess the Qt library is highly likely used with GUI stuff in almost all the cases though, unlike the C library which is also used without any GUI stuff at all, which might make this a useful thing for the Qt lib...

ximion:
I added a non-GUI icon implementation which also works with remote icons (which will be used much more in future, I assume).
We can still think about adding something QIcon based additionally though, if a GUI dependency is okay for this library...

PR #86, opened 2016-11-04
Reduce source incompatibility against last version

aleixpol:
Makes it possible to have something like this:
https://phabricator.kde.org/D3265

### DIFF
--- a/qt/CMakeLists.txt
+++ b/qt/CMakeLists.txt
@@ -76,6 +76,9 @@ install(TARGETS AppStreamQt EXPORT AppStreamLibsTarget
 
 install(FILES ${APPSTREAMQT_PUBLIC_HEADERS} ${CMAKE_CURRENT_BINARY_DIR}/appstreamqt_export.h DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/AppStreamQt")
 
+#deprecated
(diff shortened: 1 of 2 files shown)


ximion:
Is this really not solveable in the Discover branch is a reasonable way? (esp. the enum aliasing...)

ximion:
At Debian we are using this patch to build with more recent AppStream releases: https://anonscm.debian.org/git/pkg-kde/plasma/plasma-discover.git/tree/debian/patches/01_new-appstreamqt.patch
So far this seems to work well, but I haven't tested yet if this also works with older AS versions (otherwise I would have submitted it for review).
The patch is based from the one you have under review at Phab at time.

aleixpol:
Closing and moving the ifdefing to Discover.

aleixpol:
Wait, we'll still need something to find the AppStreamConfig.cmake, do you also want us to try both?

ximion:
In the patch linked above I simply tried both and emitted an error if none was found.
Not super pretty, but effective.

I very much want to get rid of the old name and API, if we re-add this to AppStream now, this means we will need to support this for a very long time again, if we don't want to break older Discover versions. By not reintroducing it, we can be sure Discover uses the new API and names when linked to the latest AppStream lib.

rdieter:
fwiw, I tried using the aforementioned debian patch on my fedora box... and plasma discover crashes pretty soon after browsing packages for a minute or 2.

ximion:
Thanks for the hint - did you try that with the current AppStream version?
I'll try a longer-running Plasma Discover session today, maybe I can reproduce this issue.

aleixpol:
I pushed it to Plasma/5.8, please test that.

PR #94, opened 2016-11-28
Make sure dependencies are fetched when looking for the package

aleixpol:


ximion:
Looks good to me - please prefix changes that affect only the Qt stuff with "qt", so I can sort them later and they can easily be grep'ed for in the logs.

PR #99, opened 2017-01-06
Don't hardcode share dir but use CMAKE_INSTALL..

heirecka:


ximion:
Is there any specific reason why you need this? E.g. only the /usr/share/metainfo directory will be scanned for metadata, and if there are distros overriding it, we would need to account for that.

heirecka:
Just an architecture based prefix /usr/arch, where the shared data still should reside in /usr/share.

Thanks for merging! Now I only need to find out what to do with appstream and my distro, other than being required for plasma-desktop ;-)

ximion:
Build data (see [appstream-generator](https://github.com/ximion/appstream-generator)) and make it available on the system somehow (as part of the archive metadata or as a package), and this should "just work".
Without the data, AppStream will only have .desktop files and data in `/usr/share/metainfo`, and knowing about software only when it is installed is pretty boring.

Issue #97, opened 2016-12-30
AppStream::Pool::load() returns false on some systems


aleixpol:
At least OpenSuse, Neon and Kubuntu (with PPAs).

Reference: https://bugs.kde.org/show_bug.cgi?id=374279

ximion:
That is expected behavior, it will return `TRUE` only if *all* data was loaded, if only one bit of data was ignored, `FALSE` is returned.
To check if there was a bigger error, the function will set a `GError` - we should expose the error message too on the Qt bindings (add a `lastError` QString property? or do something more fancy?).

Changing the meaning of the `load()` return value could break things, doing so would only make sense with some prior announcement (it would at least break appstreamcli, which uses it to show a warning message (but no hard failure) if data wasn't loaded. The function returning false is usually a hint for more experienced users to search for issues, non-technically advanced users should ignore it, since the issues is almost always in upstream software or an issue with the distributor)

PR #101, opened 2017-01-11
Provide an error message as warning output

aleixpol:
As discussed in the issue #97, there's some systems where appstream is
reporting that it doesn't load properly but the users on the platform
don't have information on how to fix it.
With this warning, they will at least have a chance to know.

### DIFF
--- a/qt/pool.cpp
+++ b/qt/pool.cpp
@@ -64,9 +64,19 @@ Pool::~Pool()
     // empty. needed for the scoped pointer for the private pointer
 }
 
-bool Pool::load()
+bool AppStream::Pool::load()
 {
-    return as_pool_load (d->m_pool, NULL, NULL);
+    return load(nullptr);
+}
+
+bool Pool::load(QString* strerror)
+{
+    g_autoptr(GError) error = nullptr;
+    bool ret = as_pool_load (d->m_pool, NULL, &error);
+    if (!ret && error) {
+        *strerror = QString::fromUtf8(error->message);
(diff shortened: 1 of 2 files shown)


qt/pool.cpp
```diff
@@ -66,7 +66,13 @@ Pool::~Pool()
 
 bool Pool::load()
 {
-    return as_pool_load (d->m_pool, NULL, NULL);
+    GError* error = nullptr;
```
ximion: Ideally use `g_autoptr` here:
`g_autoptr(GError) error = nullptr;`
(then you can get rid of the `g_error_free` line)

ximion:
I always dislike libraries printing stuff to stdout/stderr, so while I would be okay with this change, I think it's just a workaround. Can we propagate the error in a useful way somehow? (by returning an error struct, using C++ errors, or adding a lastError QString property (or struct, if we want to preserve the error code) to the class or doing something else.)

aleixpol:
I see your point, but it's not like the application can do anything other than showing the error message in a fancy pointless way.
In this case it's an error that needs to be fixed by the distribution and the distribution alone. The user shouldn't be responsible for it.

ximion:
The application using the library might want to show the information in an unobtrusive popup, add more information to it and display it, write it to its own logfile and send it to the developers with other diagnostic info and possibly other things I don't know of.
We can't know how users of libraries will use them, so giving people all important information in an easy, programmatically accessible way is key.

PR #108, opened 2017-03-20
Can't have as a mandatory dependency a deprecated system

aleixpol:
The sources aren't available by the author

fedorahosted is deprecated and there's no alternative offered by upstream.

### DIFF
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -2,9 +2,10 @@
 
 find_program (XMLTO xmlto)
 mark_as_advanced (XMLTO)
-if (XMLTO STREQUAL "XMLTO-NOTFOUND")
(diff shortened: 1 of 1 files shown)


ximion:
Just making this a warning will not make the tool not-use xmlto.
I'll need to look into what it does and how to maybe replace it (the number of tools depending on it is quite huge though, so it won't go away anytime soon).
I do wonder about the weird way this tested if the variable was set before... Not sure why this was done the way it is.

ximion:
I'll reject the PR for now, given that xmlto is under active (but very slow) development at https://pagure.io/xmlto and widely used in pretty much all distributions, and the PR will just break manpage generation on systems which don't have xmlto.

Long-term, we could think of a different way to make manpages from Docbook XML though.

PR #115, opened 2017-05-31
doc: Add "service" components and their launchables

mvollmer:


hughsie:
Out of interest, how do you see gnome-software display this type of component?

mvollmer:
> Out of interest, how do you see gnome-software display this type of component?

I was thinking that it would ignore it.

hughsie:
Works for me! :)

docs/sources/metainfo/component.xml
```diff
@@ -253,17 +253,56 @@
...
+                    <term>service</term>
+                    <listitem>
+                      <para>
+                        The software can be started, stopped, and monitored by the OS "init"
+                        facility, such as systemd. The value of the tag is a name that can be
+                        used with that facility, such as a systemd unit name.
+                      </para>
+                      <para>
+                        If more than one <literal>launchable</literal> tag of type <literal>service</literal> is
+                        present, a UI is expected to let the user manage all these services.  Thus multiple
```
ximion: Nitpick: One excessive space here...

What does "manage" mean? Do you mean start? Do you mean start/restart/stop? What happens when the service is a "oneshot" unit? Maybe clarify that.

docs/sources/metainfo/service.xml
```diff
@@ -0,0 +1,117 @@
...
+	<title>Services</title>
+
+	<section id="spec-service-introduction">
+		<title>Introduction</title>
+		<para>
+                        A service component is any software that is started and supervised by
+                        the Operating Systems "init" facility, such as systemd.
+		</para>
+		<para>
+                        A service component is any software that is started and supervised by
```
ximion: Duplicate text here.

docs/sources/metainfo/service.xml
```diff
@@ -0,0 +1,117 @@
...
+                        the Operating Systems "init" facility, such as systemd.
+		</para>
+		<para>
+			The file described in this document is built upon the generic component metadata with fields specific for services (see <xref linkend="sect-Metadata-GenericComponent"/>).
```
ximion: Don't say "file" as AppStream metadata might not be retrieved as file, and this generally is a bit confusing.
Usually, we use "component" here. Additionally, I try to avoid using the term "document" to refer to the text people are reading, because since there are XML documents and YAML documents as well as two types of AppStream "documents" this is confusing.
So, I'd probably have written something like "The component described here is built upon...."

docs/sources/metainfo/service.xml
```diff
@@ -0,0 +1,117 @@
...
+			</varlistentry>
+
+			<varlistentry id="tag-service-provides">
+			<term>&lt;provides/&gt; ↪ &lt;service/&gt;</term>
+			<listitem>
+				<para>
+					This tag is described in detail for generic components at <xref linkend="tag-provides"/>.
+				</para>
+				<para>
+					For services, at least one provided <code>&lt;service/&gt;</code> must be listed in this tag.
```
ximion: Should it? And if so, what does the `service` provided-item contain as value?

docs/sources/metainfo/service.xml
```diff
@@ -0,0 +1,117 @@
...
+			</varlistentry>
+
+                        <varlistentry id="tag-service-launchable">
+			<term>&lt;launchable/&gt;</term>
+			<listitem>
+				<para>
+					This tag is described in detail for generic components at <xref linkend="tag-launchable"/>.
+				</para>
+                                <para>
+                                        At least one launchable element with type "service" must be present.
```
ximion: Make that "service" a `<literal>service</literal>`

mvollmer:
> Will you also add the C implementation?

I'd rather not.

docs/sources/metainfo/component.xml
mvollmer: Do we need to go into that much detail here?  "Manage" should mean start/stop/restart/enable/disable/monitor, what ever the UI offers for a service.

docs/sources/metainfo/service.xml
mvollmer: Fixed.

docs/sources/metainfo/component.xml
mvollmer: Whitespace fixed.

docs/sources/metainfo/service.xml
mvollmer: Other component types also use the term "file".  Maybe we can change them all in a separate PR?

"Component" sounds wrong; we don't describe the component itself, but the format for expressing meta data about it.  So what about "component metadata format"?

docs/sources/metainfo/service.xml
mvollmer: Good point, we don't need to mandate a `<launchable>` entry.  Removed.

mvollmer:
I have pushed a new commit with the fixes. Would you like me to squash them right away into the main commit?

docs/sources/metainfo/component.xml
ximion: I usually try to be precise here, because in case someone wants to implement a UI for a new software center but doesn't have the specific knowledge about services, it's helpful to know what the spec expects from UIs.
It also helps in case someone looks at the spec at a later date and wonders what the intentions were for a specific feature.

Arguably this is a rather nit-picky case here, as in case of Unix services it's well established what you can do with them.

docs/sources/metainfo/service.xml
ximion: > Other component types also use the term "file". Maybe we can change them all in a separate PR?+

Jup, I can do that later.

> "Component" sounds wrong; we don't describe the component itself, but the format for expressing meta data about it. So what about "component metadata format"?

You are right, "component metadata format" is way better - I think just writing "metadata" is fine as well (slightly less verbose)

ximion:
You can squash and force-push, if you want :-)
But ultimately with this PR it doesn't matter, as Github allows me to do a squash-and-merge when merging it, and the PR contains a single change that goes well in one commit.

docs/sources/metainfo/service.xml
mvollmer: Done.

docs/sources/metainfo/service.xml
mvollmer: Not relevant anymore since the text has been removed.

docs/sources/metainfo/component.xml
mvollmer: We would describe the UI of the launcher here, not of the software center, no?  (For example, for desktop apps, we would be describing the GNOME Shell, not GNOME Software.)

I have tried to reword this without referring to "the UI" at all.

mvollmer:
> You can squash and force-push, if you want :-)

Okay, I'll do that!

ximion:
Looks good, thank you for your contribution!

I found some tiny indentation issues, which I'll resolve in some other commit later. Also, the C implementation is missing, but that's trivial and I'll add it for the next release. :-)

PR #117, opened 2017-06-14
Add missing getters/setters for most of the properties in AS::Component

grulja:
Note: only two of them should be missing (launchables and languages), because
we don't have wrappers for them yet

### DIFF
--- a/qt/component.cpp
+++ b/qt/component.cpp
@@ -31,6 +31,9 @@
 #include "release.h"
 #include "bundle.h"
 #include "suggested.h"
+#include "contentrating.h"
+#include "launchable.h"
+#include "translation.h"
 
 using namespace AppStream;
 
@@ -151,6 +154,31 @@ Component::~Component()
     g_object_unref(m_cpt);
 }
 
+_AsComponent * AppStream::Component::asComponent() const
+{
+    return m_cpt;
+}
+
+uint AppStream::Component::valueFlags() const
+{
+    return (uint) as_component_get_value_flags(m_cpt);
+}
+
+void AppStream::Component::setValueFlags(uint flags)
+{
+    as_component_set_value_flags(m_cpt, (AsValueFlags) flags);
+}
+
+QString AppStream::Component::activeLocale() const
+{
+    return valueWrap(as_component_get_active_locale(m_cpt));
+}
+
+void AppStream::Component::setActiveLocale(const QString& locale)
+{
+    as_component_set_active_locale(m_cpt, qPrintable(locale));
+}
+
 Component::Kind Component::kind() const
 {
     return static_cast<Component::Kind>(as_component_get_kind (m_cpt));
@@ -159,6 +187,16 @@ Component::Kind Component::kind() const
 void Component::setKind(Component::Kind kind)
 {
     as_component_set_kind(m_cpt, static_cast<AsComponentKind>(kind));
+}
+
+QString AppStream::Component::origin() const
+{
+    return valueWrap(as_component_get_origin(m_cpt));
+}
+
+void AppStream::Component::setOrigin(const QString& origin)
+{
+    as_component_set_origin(m_cpt, qPrintable(origin));
 }
 
 QString Component::id() const
@@ -184,6 +222,23 @@ void Component::setDataId(const QString& cdid)
 QStringList Component::packageNames() const
 {
     return valueWrap(as_component_get_pkgnames(m_cpt));
+}
+
+void AppStream::Component::setPackageNames(const QStringList& list)
+{
+    char **packageList = stringListToCharArray(list);
+    as_component_set_pkgnames(m_cpt, packageList);
+    g_strfreev(packageList);
+}
+
+QString AppStream::Component::sourcePackageName() const
+{
+    return valueWrap(as_component_get_source_pkgname(m_cpt));
+}
+
+void AppStream::Component::setSourcePackageName(const QString& sourcePkg)
+{
+    as_component_set_source_pkgname(m_cpt, qPrintable(sourcePkg));
 }
 
 QString Component::name() const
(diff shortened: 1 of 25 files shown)


qt/bundle.h
```diff
@@ -40,6 +40,11 @@ class APPSTREAMQT_EXPORT Bundle {
         Bundle& operator=(const Bundle& bundle);
         bool operator==(const Bundle& r) const;
 
+        /**
+         * \returns the internally stored AsBundle
+         */
+        _AsBundle *asBundle() const;
```
ximion: Are these really necessary? I would argue that this is only needed in case the Qt/C++ interface is incomplete and someone therefore wants to use the C library directly. In that case, we should rather fix the Qt binding to include the missing stuff, and not give library users the means to hack around it and not report the issue.

ximion:
Looks good on a first quick look at it, except for the obvious CI failure (which hints at an actual bug ^^).

Thank you for working on this, this is great help!

qt/component.h
```diff
@@ -138,20 +178,46 @@ Q_GADGET
...
+
+        MergeKind mergeKind() const;
+        void setMergeKind(MergeKind kind);
 
+        QHash<QString,QString> custom() const;
+        QString customValue(const QString& key);
+        bool insertCustomValue(const QString& key, const QString& value);
+
+        /**
+         * TODO contentRating, launchable and languages
```
ximion: Adding launchable to the Qt interface is actually a must-be-present feature for the next release, as that's a key new feature and using `desktopId` is deprecated in favor of it.

qt/bundle.h
grulja: I use these internally, i.e. in as_component_add_icon() or as_component_add_screenshot() and so on, because those methods takes AsIcon or AsScreenshot only.

qt/component.h
grulja: Ok, I will add them later on. I'm now working on AS::Metadata class needed for parsing.

grulja:
Crap, I didn't realize that pushing more commits will automatically add them to this review. Is is possible to somehow select which commits you want to have part of this review?

qt/component.h
ximion: OK, awesome! I was going to do this when I'm done with the parser refactoring, but it looks like that will still tle.

qt/bundle.h
ximion: Then we should make the symbols internal and hide them from the public API... The tricky bit is that they're class members...
(I will take a look on whether it's possible to even have them as internal members, probably the only way to do that is to make them private and use `friend` classes for access. - in any case, just leave this as it is for now :-) )

qt/component.h
grulja: I've implemented above mentioned missing getters/setters including new classes needed for that.

ximion: Nice! Now, fixing the CI failure would be useful, otherwise a lot of code won't work ^^

grulja: Fixed, sorry, I haven't seen the error before. I saw CI failure at the beginning but there was a different problem (something with downloading or so) and I ignored it since then.

qt/component.cpp
```diff
@@ -394,11 +560,109 @@ QList<AppStream::Suggested> AppStream::Component::suggested() const
...
 }
 
+void AppStream::Component::addSuggested(const AppStream::Suggested& suggested)
+{
+    as_component_add_suggested(m_cpt, suggested.suggested());
+}
+
+QStringList AppStream::Component::searchTokens() const
+{
+    valueWrap(as_component_get_search_tokens(m_cpt));
```
ximion: The CI complains about a missing return statement here...
And I was also surprised that this thing is public API, but looking at `AsComponent` it indeed is...

grulja: Fixed that one too, I wonder why I don't get a compilation error here while CI obviously reports that.

ximion: Configure AppStream with `-DMAINTAINER=ON`, that will enable (almost) all compile-time maintainer checks.

ximion: Probably useful to do that locally, as the CI still isn't happy :P

grulja: CI is now happy, want me to squeeze all these commit into one or later on once you are happy about that?

ximion:
Looks good to me! Can you maybe fold in the two return-fix commits, and prefix the changes with "qt"?

I am not really happy about exposing the raw GObjects in the Qt bindings in a public interface, but we can find a solution for that after merge, if there is one.

qt/contentrating.cpp
```diff
@@ -0,0 +1,148 @@
...
+    { ContentRating::RatingValueNone, QLatin1String("none") },
+    { ContentRating::RatingValueMild, QLatin1String("mild") },
+    { ContentRating::RatingValueModerate, QLatin1String("moderate") },
+    { ContentRating::RatingValueIntense, QLatin1String("intense") }
+    }
+));
+
+AppStream::ContentRating::RatingValue AppStream::ContentRating::stringToRatingValue(const QString& ratingValue)
+{
+    if (ratingValue == QLatin1String("none")) {
```
aleixpol: Why not a hash here?

qt/contentrating.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    AsContentRating *contentRating() const
+    {
+        return m_contentRating;
+    }
+
+    AsContentRating* m_contentRating;
+};
+
+typedef QHash<ContentRating::RatingValue, QString> RatingMap;
```
aleixpol: A vector should be enough?

qt/launchable.cpp
```diff
@@ -0,0 +1,130 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "appstream.h"
+#include "launchable.h"
+
+#include <QDebug>
```
aleixpol: include possibly not needed?

qt/contentrating.cpp
grulja: Not sure if this would make any difference, QHash is used accross appstream same way so I used it here as well.

qt/contentrating.cpp
grulja: Fixed to use hash even here.

qt/launchable.cpp
grulja: Removed.

grulja: Taking back, it's needed.

qt/contentrating.cpp
ximion: What's the overhead of a `QHash`? Would using a vector be much better? If not, keeping the QHash is fine, I guess.

ximion:
Okay, merged - it's one massive commit now, maybe breaking this down would have been nicer, but well, too late...

Thank you for your patch!

PR #119, opened 2017-06-20
doc: Add guidelines for distributors

jbicha:
Debian and Ubuntu have a usability problem with the GNOME Software app.

1. Removing GNOME Books removes GNOME Documents since they are both included in the same package.

2. Enabling one addon for Eye of GNOME installs all of eog's addons and disabling one addon removes all of eog's addons. (This affects a few other apps too.)

This pull request helps to fix this problem by standardizing a few best practices for distros.

The commented line looks like a typo introduced in https://github.com/ximion/appstream/commit/40431c63e7cd13f1d18f69c69e29cd0fd3f20525 so I fixed that too.

### DIFF
--- a/docs/sources/quickstart/desktop-apps.xml
+++ b/docs/sources/quickstart/desktop-apps.xml
@@ -14,7 +14,7 @@
 			For most users it allows them to answer the question <quote>Do I want to install this application?</quote>.
 			Traditionally in Linux distributions, we have none of this data for the vast majority of our desktop user-installable applications.
 			The packages-descriptions are describing all contents of a package, and not just a single application. They are also often written in a technical
-<!-- 			language and refer to other packages, which makes it hard for beginners to understand what the application they want to install really does. -->
+			language and refer to other packages, which makes it hard for beginners to understand what the application they want to install really does.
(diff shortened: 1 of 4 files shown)


docs/sources/quickstart/desktop-apps.xml
```diff
@@ -14,7 +14,7 @@
 			For most users it allows them to answer the question <quote>Do I want to install this application?</quote>.
 			Traditionally in Linux distributions, we have none of this data for the vast majority of our desktop user-installable applications.
 			The packages-descriptions are describing all contents of a package, and not just a single application. They are also often written in a technical
-<!-- 			language and refer to other packages, which makes it hard for beginners to understand what the application they want to install really does. -->
+			language and refer to other packages, which makes it hard for beginners to understand what the application they want to install really does.
```
ximion: Wtf... Thanks for spotting this!

ximion:
Looks good to me. The page is a bit short at the moment, but just having it is helpful so we can add tips as soon as we notice them (in the long run we will likely want to break up the page into a few sections and use enumerations).

PR #118, opened 2017-06-15
Add AS::Metadata class used for parsing desktop and appstream metadatadata files

grulja:
Please note that this won't compile, because it depends on my previous review needing AS::Component::asComponent() for addComponent() method.

PR #120, opened 2017-06-22
Add AS::Metadata class used for parsing desktop and appstream metadata files

grulja:
Same PR as https://github.com/ximion/appstream/pull/118, I just wasn't able to squash all the commits into one as there were some other commits among them.

### DIFF
--- a/qt/CMakeLists.txt
+++ b/qt/CMakeLists.txt
@@ -35,6 +35,7 @@ set(APPSTREAMQT_SRC
     contentrating.cpp
     launchable.cpp
     translation.cpp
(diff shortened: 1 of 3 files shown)


ximion:
Still looks good :-)
With all these new enum-to-string mappings, we need to be careful in future to adjust the Qt bindings as soon as the C bindings change.
I wonder if I could make some script to have the CI test for that...

Anyway, thanks for your contribution!
The next release will be *massive*, as I also want to get the YAML parser refactoring in, so it will likely happen in about two weeks, followed by faster releases again.

PR #123, opened 2017-07-13
trivial: adjust docs to clearly state that licenses are SPDX expressions

### LINKED ISSUE #122: project_license and SPDX
I have a bit of a question regarding the SPDX usage for appstream.

The appstream documentation says about `project_license`: 

> It should be a string in SPDX format. Licenses may be combined using and and or logic.

That seems a bit ambiguous to me. "SPDX format" would be the SPDX license expression format as the identifiers themselves aren't actually following a specified rule from what I can see. That does however not seem to be what is actually expected by appstream here.

A SPDX license expression would be this: `(LGPL­2.1+ AND (GPL­2.0+ WITH Bison­exception­2.2)) OR GPL­3.0`

The example in the appstream documentation goes on to contract this with `LGPL-3.0+ and GPL-3.0+`

Which is an expression containing SPDX license identifiers but not a valid SPDX expression (SPDX expression requires upcased AND/OR), so I am guessing the intention is a custom combination syntax with SPDX license identifiers.

If so, I think the documentation ought to say

> It should be a SPDX license identifier string. Licenses may be combined using and and or logic."

That being said, however. I would like to make a case that `project_license` should, in fact, be changed to SPDX expression syntax. The current format with and/or is deficient in that it doesn't define how to combine an exception with a license. i.e. you cannot define `GPL­2.0+ WITH Bison­exception­2.2`. 
Now, one could add that but then you might as well just go implement the expression syntax as the only other difference is bracketing, which isn't too exciting but still has its uses.

Deferring to the SPDX license expression syntax would allow passing SPDX license information verbatim from other sources. Such as auto-generating appstream data from other metadata sources (which is what we do, for example, for Plasma applets as they already have different metadata). In addition to that it would fully standardize the license expressions. So, a common SPDX parser implementation could be used to parse appstream license info etc.

apachelogger:
Previously the doc was a bit ambiguous about the actual format employed
for licenses. The format expectation is the actual SPDX license expression
format as described in the SPDX specification.
Clearly state this and point to the specification URL for reference on the
format.

Additionally, while SPDX must be parsed as case-insensitive it's considered
best practice to use strict-casing in examples, so change to that.

Closes #122

### DIFF
--- a/docs/sources/collection/xmldata.xml
+++ b/docs/sources/collection/xmldata.xml
@@ -168,11 +168,11 @@
 				<listitem>
 					<para>
 						The <code>&lt;project_license/&gt;</code> tag is indicating the license of the component.
-						It should be a string in SPDX format. Licenses may be combined using <emphasis>and</emphasis> and <emphasis>or</emphasis> logic.
+						It should be a <ulink url="https://spdx.org/specifications">SPDX license expression</ulink>.
 						Possible values include:
 						<itemizedlist>
 							<listitem><para>GPL-2.0</para></listitem>
-							<listitem><para>LGPL-3.0+ and GPL-3.0+</para></listitem>
+							<listitem><para>LGPL-3.0+ AND GPL-3.0+</para></listitem>
 							<listitem><para>MIT</para></listitem>
 							<listitem><para>CC-BY-SA-2.0</para></listitem>
 							<listitem><para>...</para></listitem>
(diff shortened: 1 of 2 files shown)


ximion:
Thanks!
The old text was a remnant from when our parser didn't yet support the full license expression syntax, but that has been fixed for years now.

PR #126, opened 2017-08-01
Don't crash when an error is found and no error string is passed

aleixpol:
See backtrace:
https://bugs.kde.org/show_bug.cgi?id=382916

I fixed it in Discover, but it shouldn't be crashing anyway.


ximion:
Indeed, thanks!

PR #133, opened 2017-08-31
Only build api docs if -Ddocumentation=true

heirecka:
This matches the behaviour of the cmake build system.
Also search for gtk-doc, which is required to build the api docs and
better fail early than later during make install.

### DIFF
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -27,8 +27,6 @@ custom_target('man-appstreamcli',
     ]
 )
 
-subdir('api/')
-
 #
 # Documentation
 #
@@ -71,8 +69,11 @@ as_doc_src = [
 ]
 
 if get_option('documentation')
+    find_program('gtkdoc-scan')
     publican_exe = find_program('publican')
     python_exe = find_program('python3')
+
+    subdir('api/')
 
     build_docs_cmd = [
             python_exe,


heirecka:
Couldn't really test with -Ddocumentation=true because publican isn't packaged here (yet).

ximion:
I *think* that the API documentation was always built, precisely because a gtk-doc dependency is easier to satisfy for distros than a dependency on Publican.

ximion:
Okay, looking at the CMake things: The API docs were not rebuilt without the `-DDOCUMENTATION=ON` flag, but the prebuilt API docs were installed. With Meson, there is no prebuilt API documentation anymore, only a prebuilt copy of the spec, for those people who do not have Publican.

I do want the API docs to be present, if I hide them behind the documentation flag, distros will just not ship them (or forget to ship them), which is not ideal.
I'll need to think about it.

Just out of interest, does your distro have MkDocs[1]?

[1]: http://www.mkdocs.org/

heirecka:
> I do want the API docs to be present, if I hide them behind the documentation flag, distros will just not ship them (or forget to ship them), which is not ideal.
> I'll need to think about it.

Sure.

> Just out of interest, does your distro have MkDocs1?

No, but that does at least not hide its tarballs behind a bunch of broken links.

PR #144, opened 2017-10-18
Add the "web-application" URL type

dchristidis:
From my [initial PR](https://github.com/hughsie/appstream-glib/pull/195) on appstream-glib:

> The Telegram web application opens its [homepage](https://telegram.org/). Instead, it should open [this](https://web.telegram.org/) location. As it stands, a user has to perform unnecessary clicks each time the application is launched. Another example is Skype ([homepage](https://www.skype.com), [web app](https://web.skype.com)), which would be a wonderful addition once WebKitGTK+ can support it.
>
> I suggest adding a new, optional type `AS_URL_KIND_WEB_APP`. If this is acceptable, I will submit a patch to GNOME Software so that it prefers it over `AS_URL_KIND_HOMEPAGE` during the web application installation.

I’m hoping that the justification is reasonable and the implementation correct.


aleixpol:
The fact that it has a web-app could be circumstancial. Sounds like it should be the value of the a `<bundle type="webapp`>https://...`.

See https://www.freedesktop.org/software/appstream/docs/chap-CollectionData.html

`<url>` is usually from before deciding to adopt the software.

ximion:
This needs some discussion first. Web applications have never been properly defined in AppStream, and it has been a long-time item on my todo list.
I actually thought about using the bundle tag for this as well, because a bundle is a way of distributing the app, while the URL tag just exists for additional information on the app.
@hughsie What do you think?

hughsie:
Web apps don't really have a bundle, they're not bundled blobs, they really are just a URL and some metadata already in the AppStream XML.

ximion:
@hughsie Jup, that's my problem with using `<bundle/>` as well. But the `<url/>` tag is thought to be additional, informative weblinks that have nothing to do with the delivery method of an app. The very purpose of the `<bundle/>` tag however is to describe how software can be made to work.

So, from that standpoint is see webapp-URLs being closer to bundle than to URL. In order for this to go into AppStream properly, we would also need to determine what a webapp is exactly, and how a software center should deal with it (just open the URL in a webbrowser? or something different?).

ximion:
Actually, the intended purpose here is closest in spirit to the `<launchable/>` tag.
So, as for the spec, I propose a `<launchable type="web-url">https://example.com</launchable>` tag addition that is required for all components of type `web-application`.

As for how software centers should handle this, there are still open questions (should they use a dedicated browser? what is the difference from a web app compared to a simple hyperlink? which websites actually are apps?)

hughsie:
Launchable makes sense to me too.

hughsie:
> I also wonder how we distinguish the current web-application components

It's a different component type for sure.

ximion:
@hughsie At the moment, those are of the "service" component-type, because they essentially are services that run in the background. But let's not worry about that yet.

dchristidis:
Thank you for discussing this issue. I have submitted a new patch for which I’m hoping to receive feedback on.

I’m not sure where to state that a `launchable` tag of type `url` is required for components of type `web-application`. It looks like there’s a need for a new section in chapter 2, but I don’t feel qualified to write it.

ximion:
> I’m not sure where to state that a launchable tag of type url is required for components of type web-application. It looks like there’s a need for a new section in chapter 2, but I don’t feel qualified to write it.

Exactly. Web applications were not really supported in AppStream until now and were more of a hidden feature. It's likely time to change that, and I already started drafting the new section in the spec.

Your patch looks fine to me, thanks!

PR #165, opened 2018-01-16
Spelling: MIME type/s

comradekingu:


ximion:
This will break AppStream, because it changes tag names to invalid ones and makes identifier strings invalid. There is a reason why those strings aren't translatable.

The only one that you could change is the very last one (the only string marked as translatable).

comradekingu:
@ximion had a feeling it might. There is however one translatable string for "mimetype",
https://hosted.weblate.org/translate/appstream/translations/nb/?checksum=1c75e5693cecdb0b which is subsequently translated as variants of MIME-type in various other languages.

ximion:
Jup ;-)
Sorry for the delayed response, so much stuff to do...

PR #169, opened 2018-01-24
Fix regression from #153

gbsf:
Define location and soname for all configurations, not just Debug.


aleixpol:
That makes sense, thanks!

ximion:
Indeed! Thanks for the patch!

PR #172, opened 2018-01-30
Add install-docs meson build option.

antlarr-suse:
--install-docs=false can be used to disable installation of documentation,
which right now can't be disabled as it was possible with cmake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,14 +62,14 @@ meson <flags> ..
 ninja
 ninja test
 ```
-Possible AppStream-specific flags are:
- -Dqt=true           -- Build the Qt interface library (default: false)
- -Dvapi=true         -- Build Vala API to use the library with the Vala programming language (default: false)
- -Ddocs=true         -- Build specification and other documentation, requires Publican (default: false)
- -Dinstall-docs=true -- Install documentation (default: true)
- -Dmaintainer=true   -- Enable strict compiler options - use this if you write a patch for AppStream (default: false)
- -Dstemming=true     -- Enable support for stemming in fulltext searches (default: true)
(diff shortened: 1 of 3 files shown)


ximion:
Why would you want to disable this at all?

Conan-Kudo:
SUSE people don't like packaging documentation, generally, from what I can tell.

README.md
```diff
@@ -61,13 +61,14 @@ meson <flags> ..
...
 ninja test
 ```
-Possible AppStream-specific flags are:  
- -Dqt=true          -- Build the Qt interface library (default: false)  
- -Dvapi=true        -- Build Vala API to use the library with the Vala programming language (default: false)  
- -Ddocs=true        -- Build specification and other documentation, requires Publican (default: false)  
- -Dmaintainer=true  -- Enable strict compiler options - use this if you write a patch for AppStream (default: false)  
- -Dstemming=true    -- Enable support for stemming in fulltext searches (default: true)  
- -Dapt-support=true -- Enable integration with the APT package manager on Debian (default: false)
+Possible AppStream-specific flags are:
```
ximion: This removes end-of-line whitespaces, which are used by Markdown to denote explicit linebreaks. If I merge this patch, the resulting Markdown rendering will look terrible.

ximion: This removes end-of-line whitespaces, which are used by Markdown to denote explicit linebreaks. If I merge this patch, the resulting Markdown rendering will look terrible.

antlarr: Oops, you're right. I added the missing spaces back with another commit

ximion:
Looks good now :-)
Thank you for the patch!

PR #177, opened 2018-02-23
Implement missing constructors

aleixpol:
Fixes a crash when no deref/ref on = operator.
Makes it possible to leverage move semantics on components.


ximion:
Looks good to me!

PR #179, opened 2018-02-25
apt: escape the icon scale factor as it is using the @ character

tintou:
It's either an issue here or with APT itself, on my system I had
/var/lib/apt/lists/packages.elementary.io_appcenter_dists_xenial_main_dep11_icons-64x64%25402.tar.gz
And of course, the hidpi wasn't available because of the arobase :)


ximion:
This makes sense! Thank you for spotting the issue and fixing it!

PR #182, opened 2018-03-09
Fix infinite loop bug.

memnoth:
If the parameter, old is an empty string for some reason, conditions of both for loops will be never finished. To fix the bug, added one more condition that if oldlen is 0, loops never run.

### DIFF
--- a/src/as-utils.c
+++ b/src/as-utils.c
@@ -456,7 +456,7 @@ as_str_replace (const gchar *str, const gchar *old, const gchar *new)
 	size_t count, retlen, newlen = strlen(new);
 
 	if (oldlen != newlen) {
-		for (count = 0, p = str; (q = strstr(p, old)) != NULL; p = q + oldlen)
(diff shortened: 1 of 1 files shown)


ximion:
@memnoth Did you ever hit this issue when using AppStream? Or did you find it when reading the code?
Thank you for your patch in any case, I fixed this now slightly differently, using GString (this way, AppStream has only one function to search & replace in a string).

memnoth:
The bug had been up in this issue: https://github.com/elementary/appcenter/issues/588
This [line](https://github.com/ximion/appstream/commit/e2cfccad378751e8b402885072656a0f394810fb#diff-63311f824c294bf2694521f3eb76d5e6L979) is related to the bug.

Thank you!

ximion:
@memnoth Oooh, very interesting! The cases you never thing would happen ^^
This warrants some special mention as a bug then, as well as a faster release of 0.12.0.

PR #185, opened 2018-04-03
Check also plain language string in as_component_localized_get

### LINKED ISSUE #166: Translated metadata is not cached if locale contains country code
Using appstream 0.11.7. If LANG=xx_XX and there is no translation in the xml metadata for lang=xx_XX, appstream should use the translation for lang=xx instead. Most packages don't ship specific translations for language variants in their metadata, so this makes descriptions be always untranslated in Discover even if there are available translations in the metadata.

antonio-rojas:
If the component doesn't contain translated data for the given locale string, try the plain language string before falling back to C

Fixes #166

### DIFF
--- a/src/as-component.c
+++ b/src/as-component.c
@@ -1119,13 +1119,13 @@ as_component_localized_get (AsComponent *cpt, GHashTable *lht)
 	gchar *msg;
 	AsComponentPrivate *priv = GET_PRIVATE (cpt);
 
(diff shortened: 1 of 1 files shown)


ximion:
This can be made a bit more efficient, but I'll change that in master.
Thank you for the PR!

ximion:
This patch also contained a memory leak, but that's all resolved in master. Thanks again for debugging the issue!

PR #186, opened 2018-04-13
Allow release description translations

mclegrand:
https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html allows to provide information about releases.
This PR aims at allowing xgettext to list them as translatable strings.


ximion:
Thank you! Looks like this was completely missed a while back...

Issue #55, opened 2016-07-23
Add property about available controls


hadess:
This would be useful for games, to know whether they can be controlled through:
- joystick (maybe with a hint as to which type is best suited)
- keyboard alone
- mouse alone
- mouse + keyboard

See https://mail.gnome.org/archives/games-list/2016-July/thread.html#00002 for the original discussion

ximion:
Sounds good. We can't name this `input_method`, because this word has meaning already.
So, something like
`<input_control>mouse</input_control>`
or
`<input_type>keyboard</input_type>`
I need to think a bit more about the name and whether it makes sense to have a tag with children for this.
@hadess Knowing nothing about joysticks, what does "joystick type" mean?

In any case, I can give this a general thumbs up, but the exact details and tag names would need to be worked out.

Kekun:
@ximion The different controllers could be:
- gamepads
- arcade sticks
- lightguns
- joysticks (for flight simulator)
- steering wheels
- dance mats
- guitars
- …

ximion:
This is resolved via https://github.com/ximion/appstream/pull/265 now. Thank you for reporting this issue!

PR #192, opened 2018-06-07
Extend recommendation expressiveness spec

aleixpol:
Include 2 extra requirement types: screen size and input type.
Make it possible to specify recommendation by whether it is just
supported.

This is a proof of concept, of course. Feedback very welcome.

### DIFF
--- a/docs/sources/metainfo/component.xml
+++ b/docs/sources/metainfo/component.xml
@@ -569,6 +569,11 @@
 				component, but may display a warning before allowing it.
 			</para>
 			<para>
+				To specify that a device is being used, a component can use the <literal>supports</literal> tag. This means that the software component
+				supports a certain configuration or other item. This is useful in case the client wants to offer components that support a certain feature
+				exclusively.
+			</para>
+			<para>
 				A <literal>requires</literal> or <literal>recommends</literal> tag contains children describing the type, value and version relation
 				of the required item.
 				Each child can have a <literal>version</literal> and a <literal>compare</literal> property, to allow depending on a certain minimal
@@ -644,6 +649,34 @@
 				<programlisting language="XML"><![CDATA[<recommends>
   <memory>2048</memory> <!-- recommend at least 2GiB of memory -->
 </recommends>]]></programlisting>
+				</listitem>
+			</varlistentry>
+
+			<varlistentry>
+				<term>&lt;input/&gt;</term>
+				<listitem>
+				<para>
+					Specifies that a input capability type is supported by the hardware present in the device.
+					Options here are: keyboard, pointer, touch, tablet-tool, tablet-pad, gesture, switch
+					<!-- As listed here https://wayland.freedesktop.org/libinput/doc/latest/group__device.html#ga3110cdddce94a1df0b8a3306909c8f15 -->
+					Example:
+				</para>
+				<programlisting language="XML"><![CDATA[<recommends>
+  <input>touch</input> <!-- recommends touch input capability -->
+</recommends>]]></programlisting>
+				</listitem>
+			</varlistentry>
+
+			<varlistentry>
+				<term>&lt;screen/&gt;</term>
+				<listitem>
+				<para>
+					Specifies a display diagonal size. Should specify the value and the unit (cm or in for centimeters or inches respectively).
+					Example:
+				</para>
+				<programlisting language="XML"><![CDATA[<requires>
+  <screen compare="ge">7in</screen> <!-- requires a 7 inch display at least-->
+</requires>]]></programlisting>
 				</listitem>
 			</varlistentry>
 


aleixpol:
ping?

hughsie:
I guess I'm not sure of the bigger picture; is a `<input>touch</input>` application going to be _hidden_ on a non-touch device, or _promoted more_ but still available on a touch-capable device?

aleixpol:
> I guess I'm not sure of the bigger picture; is a <input>touch</input> application going to be hidden on a non-touch device, or promoted more but still available on a touch-capable device?

That's why we need either `requires`, `recommends` or `supports`. If we are talking about `supports` then it should be listed, if it `requires` touch, then if there is no touch it shouldn't be displayed.

Does that make sense?

hughsie:
> That's why we need either requires, recommends or supports.

Agreed, although `supported` might be clearer than `supports`

aleixpol:
We already had `<recommends><input>touch</input></recommends>`, given this, I'd say `<supports><input>touch</input></supports>` is what makes most sense.

docs/sources/metainfo/component.xml
```diff
@@ -567,6 +567,10 @@
 				<literal>recommends</literal> tag should be used. In this case, AppStream clients should allow the installation of the software
 				component, but may display a warning before allowing it.
 			</para>
+			<para>
+				To specify that a device is being used, an application can use the <literal>supports</literal> tag. This will mean that a certain requirement
```
ximion: We never talk about "applications" in the generic component spec, because the component may be anything from a font to a shared library. The spec uses "component" or "software" (or "software component") exclusively instead.

docs/sources/metainfo/component.xml
```diff
@@ -567,6 +567,10 @@
...
+				will be integrated successfully. This is useful in case the client wants to list components that support a certain feature exclusively.
```
ximion: > This will mean that a certain requirement will be integrated successfully.

That sounds super weird. Also, whether something happens successfully or not is not something the spec knows. Maybe "This means that the software component supports a certain configuration or other item.".
But I am not sure yet whether `supports` is such a great idea yet.

Btw, please write in present tense for specification changes (it described the status quo, and rarely what will be done in future).

docs/sources/metainfo/component.xml
```diff
@@ -646,6 +650,34 @@
 				</listitem>
 			</varlistentry>
 
+			<varlistentry>
+				<term>&lt;input/&gt;</term>
+				<listitem>
+				<para>
+					Specifies that a input capability type is supported by the hardware present in the device.
+					Options here are: keyboard, pointer, touch, tablet-tool, tablet-pad, gesture, switch
+					<!-- As listed here https://wayland.freedesktop.org/libinput/doc/latest/group__device.html#ga3110cdddce94a1df0b8a3306909c8f15 -->
```
ximion: Nice, we should have a list of allowed values based on the libinput input types here.

ximion:
@aleixpol @hughsie Do we really need another toplevel tag for this?
We could have
```XML
<recommends>
  <input>touch</input>
  <input>mouse</input>
</recommends>
```
just fine, couldn't we? The only issue here is that a `recommends` tag chains its items with an implicit AND, which is wrong for input methods (you rarely use multiple of those).
If we end up having `supports` it would have to be named `supports` (not `supported`) to be in line with the existing `recommends`, `requires`, `suggests`, `extends` and `provides` tags.

Also, sorry for not replying, my schedule is crazy at time.

docs/sources/metainfo/component.xml
aleixpol: That's what we have after "options here are:", no?

aleixpol:
> The only issue here is that a recommends tag chains its items with an implicit AND, which is wrong for input methods (you rarely use multiple of those).

Maybe turning `<recommends>` into an `or` could fit the bill. Or maybe the difference between `<recommends>` and `<supports>` could be the `or` logic.

> Also, this request touches parts of another feature request filed as #55

I guess #55 don't necessarily need to be libinput types?

ximion:
> Maybe turning <recommends> into an or could fit the bill.

That would be a breaking change AND inconsistent behavior with the other tags, so that's a definite no-go...

> I guess #55 don't necessarily need to be libinput types?

Jup, so maybe that needs some slightly different solution. There is some great overlap here though.

The more I think about this issue, the better I find adding a `supports` toplevel tag - as insane as having so much expressiveness (supports, recommends, requires, extends, suggests) is, it does serve a useful purpose, and `supports` is distinct in meaning from `recommends` in that the former means a component work with other components/features, but doesn't need them to be present, while the latter (`recommends`) indicates a clear preference.
So, recommending touch input is clearly different from supporting it, and recommending a higher resolution screen is different from supporting it.

So, at the moment I am leaning towards just accepting `supports`. Implementation-wise, it would just be another enum entry in `AsRelation`.

ximion:
In order to move this forward and address the most common cases, I am currently looking at the following:
* Add a `supports` tag in addition to the existig relations tags
* Add a `chassis` relation item, with allowed values:
  * workstation
  * tablet
  * handset
  * watch
  * embedded

  This will allow software centers to hide apps based on the form factor of the current machine. See the systemd machine-info chassis types for reference: https://www.freedesktop.org/software/systemd/man/machine-info.html (I folded desktop, laptop and server into "workstation" here, I still need to think whether that is a good plan)
* Defer the `input` and `screen` relation item discussion to later, to keep this simple for now. Those tags are actually quite hard to test for application authors (esp. the screen one). We can add them at a later time if we see that they are necessary.

What do you think?

cassidyjames:
>I folded desktop, laptop, convertible and server into "workstation" here, I still need to think about whether that is a good plan

Something we learned in elementary OS when designing power settings was that distinguishing based on a "group" of hardware gets increasingly arbitrary and confusing as device form factors overlap. We've found it almost always makes more sense to just define the actual hardware we require or are looking for—so in some cases it's an embedded display or attached battery, other times it's a touchscreen, sometimes it's an accelerometer. My concern with the proposed chassis groupings is: what distinguishes a "handset" from a "tablet"? What separates a "tablet" from a "convertable"? These distinctions always fall back to some specific hardware differentiation, so I don't know that creating these buckets makes sense. Especially if I have a small phone-sized tablet with a physical keyboard.

aleixpol:
Doint it around "chasis" won't solve #55 either. It's just turning into information we don't know about.
In the end, the software centers can decide it's a watch because the screen is 3cm.

Maybe using libinput as a reference was wrong because it's a very specific kind of input (was thinking voice) but libappstream can't be guessing, as cassidy puts it.

`supports` is what we have already on the patch.

ximion:
So, I started that discussion again on the AppStream ML: https://lists.freedesktop.org/archives/appstream/2020-February/000337.html - hopefully we can get some kind of solution for this.

ximion:
This is resolved via https://github.com/ximion/appstream/pull/265 for input controls. Screen sizes still need a bit more work and research.
If you have insights or feedback to give, please participate at https://lists.freedesktop.org/archives/appstream/2020-February/000337.html
Thanks to anyone involved with this PR, it was very helpful to get this ball rolling (even if it rolled reeeeeealy slowly, which was my fault).

PR #201, opened 2018-08-28
Set error variable if there were some errors during refinement

StefanBruens:
When as_pool_refine_data encounters bad components, this is reflected
in the return code, but there is no further information given what went
wrong. Give a short hint to the user, and refer to the debug output which
allows determining the cause.


ximion:
This makes sense. Thank you for the patch!

ximion:
Looks like I was a bit too quick with merging this:
```
../src/as-pool.c:1020:12: error: implicit declaration of function 'g_new_error_literal'; did you mean 'g_set_error_literal'? [-Werror=implicit-function-declaration]
   *error = g_new_error_literal(AS_POOL_ERROR,
```

StefanBruens:
oops - g_error_new_literal ...

PR #203, opened 2018-10-07
Allow multiple addons per component

davidmhewitt:
Currently, only a maximum of one addon is associated with a component.

As far as I understand it, a component should be able to have multiple addons. I.e. multiple components can all have an `extends` field that points to the same component.

In this PR, when looking at the extends field, make sure we add the current component to the addons section of the extended package if the current component isn't in the addons list already, rather than artificially limiting it to 1 addon per component.

### DIFF
--- a/src/as-pool.c
+++ b/src/as-pool.c
@@ -486,7 +486,7 @@ as_pool_update_addon_info (AsPool *pool, AsComponent *cpt)
 		}
 
 		/* don't act if we already have addons */
-		if (as_component_get_addons (extended_cpt)->len > 0)
+		if (g_ptr_array_find (as_component_get_addons (extended_cpt), cpt, NULL))
 			continue;
 
 		as_component_add_addon (extended_cpt, cpt);
--- a/src/as-pool.c
+++ b/src/as-pool.c
@@ -485,7 +485,7 @@ as_pool_update_addon_info (AsPool *pool, AsComponent *cpt)
 			return;
 		}
 
-		/* don't act if we already have addons */
+		/* don't add the same addon more than once */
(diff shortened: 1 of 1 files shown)


davidmhewitt:
For testing:
`appstreamcli get org.gnome.gedit.desktop --details`

In master (on Ubuntu 18.04), you get:
`Add-ons: gedit-dashboard (gedit)`

With this branch:
```
Add-ons: 
   - gedit-dashboard (gedit)
   - gedit-multiedit (gedit)
   - gedit-terminal (gedit)
   - gedit-synctex (gedit)
   - gedit-textsize (gedit)
   - gedit-bracketcompletion (gedit)
   - gedit-codecomment (gedit)
   - gedit-bookmarks (gedit)
   - gedit-translate (gedit)
   - gedit-commander (gedit)
   - gedit-wordcompletion (gedit)
   - gedit-charmap (gedit)
   - gedit-colorschemer (gedit)
   - gedit-joinlines (gedit)
   - gedit-colorpicker (gedit)
   - gedit-git (gedit)
   - gedit-findinfiles (gedit)
   - gedit-drawspaces (gedit)
   - gedit-smartspaces (gedit)
   - gedit-zeitgeist (gedit)
```

ximion:
Makes sense to me. This is probably not the most efficient solution to this problem, since `g_ptr_array_find` isn't very fast, however since most components have very short addon lists it's likely preferable to making a hash map.
I am also planning some rework of the refining process anyway, so we might not need to run `_find` at all in future.
For the time being, this patch is a good solution.
Thank you for submitting the patch!

PR #204, opened 2018-10-10
component: Mark relevant return types as nullable

davidmhewitt:
These types are not marked as nullable in the resulting GIR file and hence not marked nullable in any bindings downstream of that; for example, the Vala bindings.

Having these marked as nullable would allow more robust Vala code to be generated when these methods are used.

This only touches Component for now in case there's some reasoning behind why these haven't been marked `nullable`, but I'm happy to submit a PR to cover off the rest if this is wanted.


ximion:
Makes sense, I think when the code was written I didn't know that return values can be tagged as nullable as well.
Thank you for the patch!

PR #206, opened 2018-10-11
Add null checks in as_release_vercmp

davidmhewitt:
As `as_release_get_version` can return `NULL`, it would be wise to do some null checks here before passing to the compare method.

### DIFF
--- a/src/as-release.c
+++ b/src/as-release.c
@@ -268,8 +268,20 @@ as_release_set_version (AsRelease *release, const gchar *version)
 gint
 as_release_vercmp (AsRelease *rel1, AsRelease *rel2)
 {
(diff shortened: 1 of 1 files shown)


ximion:
I was coincidentally looking at this as well just  few hours ago.
My patch fixes the issue differently, in the string comparison function itself. This protects every (internal) user of the API against crashes when passing NULL to the function by accident.
The code is also smaller, g_strcmp0 handles NULL values well already - problems can only exist of only one of the version numbers passed in happens to be NULL.

Thanks for your patch anyway and for noticing the issue!

davidmhewitt:
Yeah, no worries, that makes more sense. Thanks for your feedback and fix!

Issue #155, opened 2017-12-18
more url types in appdata


hsitter:
The current URL types in the appdata spec seem a bit restrictive. They do very poorly in getting people to inter-active help venues (in particular as the current help type is meant for passive help such as documentation as per the spec).

To improve this I would like to suggest extending URL types:

- forum/helpdesk for linking to support forums and the like
- chat for link to IRC/matrix/gitter/whatevs support venues
- mailing list
- wiki? (arguably falls into the existing `help` type)

robert-ancell:
Snaps have a "contact" URL to allow developers to give an easy way for users to contact them. This could be an email address (mailto link) or a website (e.g. a form or a page of contact links). I think this would be generally useful in AppStream.

One of the downsides of having a lot of different URLs is you force the user to choose which method is appropriate to contact with and the UI displaying them to decide how to best show them. Using a single contact URL allows the developer to control this. Less URLs also are more likely to have good data filled out than many.

ximion:
Yes, I don't think adding a ton more URL types to the ones we already have is a good idea at this point.
But if you find a particular one that's still missing, please file a separate bug report for the particular one, so we can discuss adding it! :-)

PR #209, opened 2018-12-04
Add a contact URL type

robert-ancell:


robert-ancell:
Raised in https://github.com/ximion/appstream/issues/155

ximion:
Thank you for the patch!

PR #213, opened 2019-01-09
qt: expose the spdx parser to Qt users

aleixpol:


qt/spdx.cpp
```diff
@@ -0,0 +1,53 @@
...
+    return AppStream::valueWrap(as_spdx_license_tokenize(qPrintable(license)));
+}
+
+QString AppStream::SPDX::detokenizeLicense(const QStringList &license_tokens)
+{
+    gchar** tokens = AppStream::stringListToCharArray(license_tokens);
+    return QString::fromUtf8(as_spdx_license_detokenize(tokens));
+}
+
+QString AppStream::SPDX::asLicenseId(const QString &license)
```
ximion: This should likely be called `toSpdxId` or `licenseToId`, since the purpose of this function is to create a SPDX-compatible identifier from a more "generic" license name.

ximion:
Thank you for the patch!

PR #214, opened 2019-01-16
as-pool: Fix g-i for get_components_*() while GPtrArray owns its elements

ricotz:
See 8213577d1ee72385a27c79a546564a52bcfba93a


ximion:
Woah, copy-paste error as it looks like.
Thanks!

PR #215, opened 2019-01-29
doc: help distinguish the 2 'languages' tags

Beuc:
Reference Collection's 'languages' tag from the localization page, so upstream don't get confused into generating language packs metadata to reference bundled gettext .mo files. Both pages define a 'languages' tag.


Beuc:
Ping? :)

ximion:
Thank you for the patch and sorry for the delay!

Issue #222, opened 2019-03-03
0.12.6 test failures on Ubuntu


jbicha:
appstream 0.12.6 fails to build on Ubuntu because of test failures:

https://launchpad.net/ubuntu/+source/appstream/0.12.6-1

ximion:
Oh wow, what happened there? Nearly every test that reads or writes AppStream data failed.
The same code tests fine on the CI, and also ran fine on Debian during the build process.
The weirdest thing is that XML results are somehow suddenly showing YAML output and vice versa.

ximion:
I've fixed a potential diff output confusion when tests fail in parallel, but that's not the cause of this problem here.
Can you reproduce the issue locally?

jbicha:
Yes, more specifically, it looks like the issue is triggered by glib 2.59.

ximion:
Hmm, from the changelog there at least is no super obvious change in GLib that could have triggered this.

jbicha:
Does this help?

https://gitlab.gnome.org/GNOME/tracker/commit/c58f7aa419e2

ximion:
Ah yeah, that change... - I was actually quite confident AppStream wasn't depending on GHashTable ordering, but that actually would explain the test failures.
I'll have to run the tests locally with a recent GLib though to find all the problems.

robert-ancell:
https://github.com/ximion/appstream/pull/223

ximion:
Those are fully fixed for a while now, and will soon be in a long overdue release.

PR #223, opened 2019-03-22
Fix tests breaking in GLib 2.59 (GHashTable order changed)

robert-ancell:


robert-ancell:
From https://github.com/ximion/appstream/issues/222

robert-ancell:
This doesn't fix all the code that uses the ordering, but seems to be enough to fix the tests.

ximion:
Looks good to me, thank you for doing all the work!
I also like that the sorting overhead ended up almost exclusively in the serialization functions (I don't think this will work forever though ^^).
The changes should be fairly complete, given that one of the tests involves loading a large XML file, serializing it to the GVariant cache, loading it and then comparing the result.
But any further issues will be uncovered by future testsuite expansions :-)

PR #226, opened 2019-04-12
Add a note about not dropping .desktop suffix from existing IDs

kalev:


TingPing:
Couldn't you just note to add the old one to `provides`?

ximion:
Yes, I will merge this later and change the wording a bit to include more information. Also, maybe putting it into a `note` box makes sense for increased visibility.

kalev:
Great, thanks!

PR #228, opened 2019-04-18
docs: Explicitly rule out nested lists

mvollmer:
Cockpit doesn't support those, but we have a request for it. Let's clarify the spec.

If nested lists are in fact supported, that's just fine as well, of course, but let's clarify that also.


hughsie:
Nested lists would also make gnome-software sad.

ximion:
Indeed!
Thanks for the patch!

Issue #229, opened 2019-04-23
Need an option not to extract component name with xgettext


fujiwarat:
Recently I migrated ibus-anthy to use xgettext from intltool.
And I noticed the translatable strings are extracted by metainfo.its.

I don't want to extract the component name because translators could translate the branding name by mistake.
I need the option not to extract the component name.

fujiwarat:
Great!

PR #230, opened 2019-04-23
its: Add translatable option in metainfo.its

fujiwarat:
Some package does not wish to extract the component name
as translatable since it's the branding name and add the
translatable option.

https://github.com/ximion/appstream/issues/229


ximion:
Looks good to me, but why would you ever not want to translate an agreement section name?

fujiwarat:
Sorry, you're right. I deleted the agreement section name line.

ximion:
Makes sense to me now! Thank you for the patch!

Issue #235, opened 2019-05-15
Improve <translation/> description in spec


kparal:
After the discussion in https://github.com/hughsie/appstream-glib/issues/302, I have a few suggestions how to improve `<translation/>` description in AppStream spec:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#sect-Metadata-GenericComponent

1. It helps a lot if the purpose of this tag is explained even to non-programmer people (like packagers). I suggest changing this sentence:
> It may be used by the AppStream distro metadata generator to determine the translation status of the respective software. 

to something like this:
> It may be used by the AppStream distro metadata generator to determine the translation status of the respective software (e.g. which languages the software is translated into and how complete the translations are).

2. Similarly to 1), I didn't understand the term "translation domain" and I didn't manage to find useful help online. In the end I realized it's the name of the `*.mo` file (for gettext) and then it's clear. It would be helpful it that could be explained better somehow (for all translation systems), but I don't know how.

3. @hughsie said `appstream-glib` now supports `.pak` and `.xpi` translation files. I'm not sure about your reference implementation, but is it something that should be incorporated in the spec as well?

4. It is unclear that `<translations/>` (an empty tag) is valid and what exactly it means. It can mean "this software is not translated" or it can mean "the translation status can't be determined by our tooling". (Now that appstream-glib made this tag mandatory in default validation pass, this is a question that will be asked by every packager of either proprietary software or software using unsupported translation systems). It would be good to clarify this, and ideally support both meanings in some way.

Thanks for consideration.

ximion:
To 1): PR to change the text is very welcome :-)

To 2): The term "translation domain" is pretty common and should be well known to people using Gettext. I would expect these people to write the initial metainfo file as well, packagers should usually not need to know about this. But if the wording could be made more clear, patches are welcome there as well.

To 3): I rarely include stuff in the spec which is only supported by one implementation. The problem with translations made with other systems is that unlike Gettext those usually have no default paths to search for translations in, so hunting them down is really hard. So far we haven't found proper solutions for this issue.

To 4): I am quite certain that `appstreamcli validate` will currently see an empty `translation` tag as a bug and fail validation, as its meaning is undefined. If it was defined, its meaning would be "this software is untranslated" as that's what empty AppStream tags generally mean in other places (that and "assume default values").

kparal:
> To 1): PR to change the text is very welcome :-)

OK, please see #236.

[...]

PR #236, opened 2019-05-17
spec: clarify <translation/> purpose

kparal:
Related: https://github.com/ximion/appstream/issues/235


ximion:
Thank you for the patch!

PR #241, opened 2019-06-23
Spelling: DBus→D-Bus, MIME types → Media types

comradekingu:
As per 
https://en.wikipedia.org/wiki/D-Bus
and
https://www.iana.org/assignments/media-types/media-types.xhtml


ximion:
Interesting that they renamed MIME Types recently...
I guess having "Media types" as user-facing description string instead makes sense, because it's much more obvious to people on what it does.
Thanks for the patch!

PR #249, opened 2019-08-28
typo replace "Manu" with "Menu" in...

leela52452:
...text: The category defined in the .desktop file is not valid.
Refer to the XDG Manu Specification for a list of valid categories.


ximion:
Thank you for the PR! :-)

PR #262, opened 2019-12-11
Use the same locale fallback for releases as for components

tintou:
This fixes the issue with release metadata not showing the translation in elementary AppCenter

### DIFF
--- a/src/as-release.c
+++ b/src/as-release.c
@@ -194,6 +194,35 @@ as_release_class_init (AsReleaseClass *klass)
 }
 
 /**
+ * as_release_localized_get:
+ * @release: a #AsRelease instance.
+ * @lht: (element-type utf8 utf8): the #GHashTable on which the value will be retreived.
+ *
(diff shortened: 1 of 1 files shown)


tintou:
My locale is `fr_FR.UTF-8` but we only provide translations for `fr` not `fr_FR`

ximion:
Hey, just a quick feedback: Your patch fixes this issue only for one particular locale, but I think this has to be addressed in a much more generic way in a different place (AsContext maybe) and with less code duplication.
I am keeping this PR open at time as a reminder to work on this, and I'll merge it when I start working on the generic solution.
Originally I wanted to do one more AppStream release before the end of the year, but it appears like we'll instead get one with this issue fixed at the beginning of January next year.
Thanks for your contribution and sorry for not replying sooner!

ximion:
I've implemented this differently now, but also in a more complete way, so the same algorithm is now used for every translation that AppStream has.
Thank you for the initial PR!

PR #265, opened 2020-02-22
User input control information in AppStream

ximion:
Hi!
These are the proposed spec changes and an initial implementation (missing tests and specific validator checks at the moment) of **user input controls**.
These can be used to indicate which methods of control by a user an application supports. This permits filtering applications that are mobile-ready to a limiting degree by checking whether they support touch input. Also, games could mark that they do support game controllers, as initially requested in https://github.com/ximion/appstream/issues/55. Applications that do have a CLI or support voice/vision as methods of user control can also be pointed out to the user.

This PR implements the addition proposed as https://lists.freedesktop.org/archives/appstream/2020-February/000339.html which also contains additional information about its rationale.
Feedback is welcome!

### DIFF
--- a/docs/sources/metainfo/component.xml
+++ b/docs/sources/metainfo/component.xml
@@ -789,6 +789,48 @@
 				</listitem>
 			</varlistentry>
 
+			<varlistentry id="tag-requires-recommends-control">
+				<term>&lt;control/&gt;</term>
+				<listitem>
+				<para>
+					This item type can be used to recommend or require certain ways a user can control the software. This usually maps to certain methods
+					of input. If multiples of these tag are found within a requires/recommends block, only one of them needs to be satisfied on the system
+					to mark an application as compatible. This means if <literal>touch</literal> and <literal>pointing</literal> are both recommended as controls,
+					an system that only has a mouse and no touchscreen will still be considered able to run the application.
+					Valid values for this tag are:
+				</para>
+				<itemizedlist>
+					<listitem><para><code>pointing</code> - Input via mouse/cursors/other pointing devices is possible</para></listitem>
+					<listitem><para><code>keyboard</code> - Keyboard input is possible</para></listitem>
+					<listitem><para><code>console</code> - Control via a console / command-line interface</para></listitem>
+					<listitem><para><code>touch</code> - Input by touching a surface with fingers is possible</para></listitem>
+					<listitem><para><code>gamepad</code> - The component supports gamepads (any game controller with wheels/buttons/joysticks)</para></listitem>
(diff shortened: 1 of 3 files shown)


aleixpol:
+1 makes sense

docs/sources/metainfo/component.xml
```diff
@@ -789,6 +789,48 @@
 				</listitem>
 			</varlistentry>
 
+			<varlistentry id="tag-requires-recommends-control">
+				<term>&lt;control/&gt;</term>
+				<listitem>
+				<para>
+					This item type can be used to recommend or require certain ways a user can control the software. This usually maps to certain methods
```
hughsie: Which is it, _recommend_ or _require_?

hughsie: maybe just change the wording to `and`?

hughsie:
No concerns from me.

docs/sources/metainfo/component.xml
ximion: It's either recommend or require, depending on which of the two tags the item is declared in. I think and is actually equally valid to "or" here, I'll have a look at the rest of the specification and change all of these occurrences in one go to what makes the most sense (never really thought about it that much).

ximion:
Neat :-)

docs/sources/metainfo/component.xml
```diff
@@ -789,6 +789,48 @@
...
+					<listitem><para><code>console</code> - Control via a console / command-line interface</para></listitem>
+					<listitem><para><code>touch</code> - Input by touching a surface with fingers is possible</para></listitem>
+					<listitem><para><code>gamepad</code> - The component supports gamepads (any game controller with wheels/buttons/joysticks)</para></listitem>
+					<listitem><para><code>voice</code> - The software can be controlled via voice recognition/activation</para></listitem>
+					<listitem><para><code>vision</code> - The software can be controlled by computer vision / visual object and sign detection</para></listitem>
+				</itemizedlist>
+				<para>
+					If a control type is <emphasis>recommended</emphasis>, it means the software supports the given method of user input. As long as one of the input methods
+					is available on the system, the software can be used. Installation on systems without the given control is still permitted.
+					If a control type is <emphasis>required</emphasis>, the same applies, but the software installed should refuse to install the application on devices which
```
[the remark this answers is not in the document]
ximion: Typo indeed - fixed via https://github.com/ximion/appstream/commit/ee40deb644007b115fc4f255c482bc9872ed1632

Btw, the experimental metainfo-creator on https://www.freedesktop.org/software/appstream/metainfocreator/ should already produce valid new-style XML :-)

PR #275, opened 2020-06-01
qt: add missing Provided::KindId enum

aleixpol:
It was never added to the Qt interface


ximion:
Indeed looks like that was missed. Thanks for the patch!

PR #276, opened 2020-06-12
docs: Update to remove icon tag from Metadata

Siosm:
This is no longer accepted by appstream-util validate-relax:

$ appstream-util validate-relax org.kde.okteta.appdata.xml
org.kde.okteta.appdata.xml: FAILED:
• tag-invalid           : <icon> not allowed in desktop appdata
• tag-invalid           : stock icon is not valid [okteta]
Validation of files failed

See also discussion in https://github.com/keepassxreboot/keepassxc/issues/1131

### DIFF
--- a/docs/xml/metainfo-component.xml
+++ b/docs/xml/metainfo-component.xml
@@ -157,30 +157,6 @@
 		</listitem>
 		</varlistentry>
 
-		<varlistentry id="tag-icon">
-		<term>&lt;icon/&gt;</term>
-		<listitem>
-			<para>
-				The <code>&lt;icon/&gt;</code> tag describes the component icon. It is mostly used for GUI applications (component-type <literal>desktop-application</literal>).
-				It can be of type <literal>stock</literal>, <literal>local</literal> or <literal>remote</literal>.
-			</para>
-			<para>
-				<literal>stock</literal> icons are loaded from the icon stock (the current or hicolor/locolor fallback themes).
-				The icon name must not include any file-extension or path.
-			</para>
-			<para>
-				<literal>local</literal> icons are loaded from a file in the filesystem.
-				They should specify a full file path.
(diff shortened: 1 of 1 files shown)


ximion:
`appstream-util` is wrong and not spec compliant, Please get that fixed in `appstream-util` or validate with `appstreamcli` instead or just remove the explicit icon tag from your metainfo file and have the AppStream-parsing tools fetch the icon from an associated .desktop file.

travier:
Will do. Thanks.

ximion:
> appstream-util is wrong and not spec compliant[...]

Just to avoid any misunderstanding: Not spec compliant *in this instance* (some apps like WINE actually rely on that feature, and tools which generate their .desktop files from metainfo files via `appstreamcli make-desktop-file` do too, so do projects who want to override a .desktop-file-specified icon in metainfo files for some reason (I only saw that particular case once though)).
This patch would have removed support for icons from the specification entirely, for all component types, btw, which is why CI failed ;-)

travier:
I found https://github.com/hughsie/appstream-glib/issues/360 & https://github.com/hughsie/appstream-glib/issues/243 for reference for anybody looking for more history.

sergiomb2:
https://www.freedesktop.org/software/appstream/docs/chap-CollectionData.html#sect-AppStream-XML

this documentation still outdated ?

ximion:
What do you mean? The documentation was always correct on this, AFAIR.

sergiomb2:
sorry I though I was point to icon section https://www.freedesktop.org/software/appstream/docs/chap-CollectionData.html#tag-ct-icon 

in icon section document says that <icon type="stock"> is valid when apparently is not.

ximion:
It is and always has been valid though. `appstream-glib` simply had a bug.

PR #284, opened 2020-10-14
Use glib-mkenums to create the GType of all the available enums

tintou:
This allows the GObject Introspection to correctly understand the relations between all the from_string/to_string methods and the enum they belong to.

This also allows to use the right GType when dealing with flags or enums.


ximion:
Is this amount of hard-coding really necessary? Is there a bug report against gobject-introspection, so, will this issue be fixed upstream in any reasonable amount of time?
This change will also break the existing Vala API, right? Or have the methods been unavailable to Vala before?

tintou:
@ximion I indeed found the root cause of this issue: none of these enums were registered to the GObject Type system, I've added the use of glib-mkenum to generate their GType.

ximion:
Looks like this got lost in ancient times during the cmake transition (!). Great find, and the patch looks good to!
Thanks for looking into this!

Issue #285, opened 2020-10-20
Allow components to specify preferred display properties for better mobile device support


ximion:
Hi!
It is useful for software centers and shells to determine whether a (graphical) application is suited for specific screen sizes as well as potentially screen resolutions. One usecase is for software centers to only show applications suitable for a phone device when run on a mobile phone, and only ones suitable for a TV screen when run on a smart TV.

### History
In the past I wanted to just encode the chassis type as defined by systemd's [machine-info](https://www.freedesktop.org/software/systemd/man/machine-info.html) specification. In prior discussions, this idea was shot down by multiple partys though, because it is very inflexible and we don't really have a clear definition of what a "phone" actually is. We could define that, of course, but our definition would always be insufficient for new, emerging device types. Who would have thought that smart watches would be a thing so soon as well? In addition to that, a lot of apps are adaptive now and run on ranges of screen sizes with diferent user input methods, and we need to account for that as well.

To solve this, I proposed a mechanism based on capabilities, which raised no complaints, except for the screen dimension approach, which we deferred at the time as "a problem to solve later". Well, I think the time to address this is now. Not only do we want this feature in Purism as we happen to develop a mobile phone, but there are also multiple other phone and "Linux on other devices than a PC" projects that I am aware of that have asked for it.

Support for the user input control property has already been implemented (see https://github.com/ximion/appstream/pull/265) and is widely available now, so the only missing bit for apps to advertise support for different device types is information about the screen (I am aware that some people also would like to specify support for additional peripherals, but that's - again - a problem for later ;-) ).

### Prior art

To move this effort forward and not have it get stuck for a while like before, I looked at some prior art. First, CSS apparently had a [media-types](https://drafts.csswg.org/mediaqueries/#media-types) query which is deprecated now in favor of the device-feture-based [media features](https://drafts.csswg.org/mediaqueries/#mq-features) tests.
This makes me feel very good about not using the chassis typy in AppStream but to go for a feature-based approach as well.

Android has a [way to set restrictions on screen properties](https://developer.android.com/guide/practices/screens-distribution) in a manifest file as well, and this one is pretty close to something we can use in AppStream. Android uses sizes like "small", "large", etc. for screens, which actually resolve to dimensions in [dp](https://en.wikipedia.org/wiki/Device-independent_pixel).

[...]

PR #286, opened 2020-10-31
WIP: Specify display_length requires/recommends item

ximion:
This change adds specification text to allow applications to require a minimum display width and/or height in order to be installed or run with recommended settings.
This change would address https://github.com/ximion/appstream/issues/285 

CC: @hughsie @apol @hadess @Exalm

### DIFF
--- a/docs/xml/metainfo-component.xml
+++ b/docs/xml/metainfo-component.xml
@@ -834,6 +834,89 @@
 				</listitem>
 			</varlistentry>
 
+			<varlistentry id="tag-requires-recommends-display-width-height">
+				<term>&lt;display_length/&gt;</term>
+				<listitem>
+				<para>
+					Set a relation to the display length defined as an integer value in either <emphasis>millimeters</emphasis> or <emphasis>logical pixels</emphasis>
+					(device pixels divided by scaling factor, roughly equivalent to 0.26mm (1/96in), also known as device-independent pixels).
+					To use millimeters, the <literal>unit</literal> property of this tag needs to be set to value <code>mm</code>, to use logical pixels
+					its value must be <code>px</code>. If no unit is explicitly defined, <code>px</code> is implicitly assumed.
+					Setting the <literal>side</literal> property to either <code>shortest</code> or <code>longest</code> will apply the selected size constraint to
+					either the shortest or longest side of the display rectangle, with <code>shortest</code> being implicitly assumed if no value is set.
+				</para>
+				<note>
+					<title>About Pixel Dimensions</title>
+					<para>
+						One logical pixel (= device independent pixel) roughly corresponds to the visual angle of one pixel on a device with a pixel density of
+						96dpi and a distance from the observer of about 52cm, making the physical pixel about 0.26mm in size.
+						If using logical pixels as unit, they might not always map to exact physical lengths as their exact size is defined by the device providing
+						the display.
+						They do however accurately depict the maximum amount of things that can be drawn in a direction on screen.
+					</para>
+				</note>
+				<para>
+					Relations for the display length can be defined using a <literal>compare</literal> property as described in <xref linkend="tag-requires-recommends"/>.
+					If this property is not present, a value of <code>ge</code> (greater-or-equal) is implicitly assumed.
+				</para>
+				<para>
+					The <literal>display_length</literal> tag also accepts one of the following text values. While their exact meaning in terms of pixel-based or physical size
+					is implementation-defined, the text term must roughly match the screen size of the device class listed next to it in the listing below:
+				</para>
+				<itemizedlist>
+					<listitem><para><code>xsmall</code> - Very small screens, as used in watches, wearables and other small-display devices (about &lt;= 300px).</para></listitem>
+					<listitem><para><code>small</code> - Small screens often used in handheld devices, such as phone screens, small phablets (about &lt; 600px).</para></listitem>
+					<listitem><para><code>medium</code> - Screens in laptops, tablets (about &gt;= 600px)</para></listitem>
+					<listitem><para><code>large</code> - Bigger computer monitors (about &gt;= 1024px)</para></listitem>
+					<listitem><para><code>xlarge</code> - Television screens, large projected images</para></listitem>
+				</itemizedlist>
+				<para>
+					If a text value is used, the <literal>unit</literal> and <literal>side</literal> properties must not be present. For <literal>side</literal>, <code>shortest</code>
+					is assumed in this case.
+					A <literal>compare</literal> property is permitted and will compare the text placeholder values from smallest (<code>xsmall</code>) to largest (<code>xlarge</code>).
+					The text values are intended for adaptive applications which only need or want to give a very rough hint as to which display lengths they support, and do
+					not need fine control over their visibility (as these types of applications will adjust well to most screen sizes at runtime).
+					If finer control is needed, absolute sizes should be used instead.
+				</para>
+				<para>
+					This tag may appear up to four times to set a minimum and maximum dimension required. It is recommended to not mix different units
+					(so e.g. a required minimal width is not requested in both <code>px</code> and <code>mm</code>).
+					If multiple displays are connected to a device, it is acceptable to test against either the largest screen attached to the device, or the combined
+					amount of display space (depending on what makes sense for the respective device).
+				</para>
+				<para>
+					If used in a <literal>requires</literal> block, this relation can be used to restrict an application to only be installable on systems which have a minimum
+					usable display length available for it. If used in a <literal>recommends</literal> block, the application will still be
+					installable, but the user may be shown a warning.
+				</para>
+				<para>
+					If no <literal>display_length</literal> relation is present, a minimum required display (<code>ge</code>) relation
+					of <code>medium</code> is implicitly assumed to preserve backwards compatibility (so applications capable of running on smaller screens
+					need to make their support for that configuration explicit).
+				</para>
+				<para>
+					Examples:
+				</para>
+				<programlisting language="XML"><![CDATA[<!-- recommend at least 600 logical pixels of space -->
+<recommends>
+  <display_length compare="ge">600</display_length>
+</recommends>
+
+<!-- require a device with at least 8x5cm physical screen space for this app -->
+<requires>
+  <display_length side="longest" unit="mm" compare="ge">80</display_length>
+  <display_length side="shortest" unit="mm">50</display_length>
+</requires>
+
+<!-- ensure this application is not run on a very large screen, or
+     very small screen (no tiny handhelds or television screens) -->
+<requires>
+  <display_length compare="lt">xlarge</display_length>
+  <display_length compare="gt">xsmall</display_length>
+</requires>]]></programlisting>
+				</listitem>
+			</varlistentry>
+
 			</variablelist>
 		</listitem>
 
--- a/docs/xml/metainfo-component.xml
+++ b/docs/xml/metainfo-component.xml
@@ -807,6 +807,7 @@
 					<listitem><para><code>console</code> - Control via a console / command-line interface</para></listitem>
 					<listitem><para><code>touch</code> - Input by touching a surface with fingers is possible</para></listitem>
 					<listitem><para><code>gamepad</code> - The component supports gamepads (any game controller with wheels/buttons/joysticks)</para></listitem>
+					<listitem><para><code>tv-remote</code> - Input via a TV remote (with arrow keys, number pad, other basic inputs) is supported.</para></listitem>
 					<listitem><para><code>voice</code> - The software can be controlled via voice recognition/activation</para></listitem>
 					<listitem><para><code>vision</code> - The software can be controlled by computer vision / visual object and sign detection</para></listitem>
 				</itemizedlist>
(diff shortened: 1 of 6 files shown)


ximion:
One alternative way of writing this, in case we do not care about both width/height that much, would be
```xml
<display_length side="shortest" unit="cm">10</display_length>
```
to only set constraints on the shortest/longest edge of a device's screen. This may be enough for most devices (shortest side being default if `side` isn't given as property) and would avoid having special wording for devices with screen rotation.

agx:
My take away from the discussion so far was: logical pixel is must have. It's currently mostly interesting for small screen sizes but might also be useful for large screens to avoid lots of unused space.

Physical size + viewing distance would be important for e.g projectors and TVs but it's orthogonal to logical pixels so i'd not try to lump both into one but make them clearly separate things to make sure they don't get confused,

bertob:
My current theory is that logical pixel size is the only thing we need/want from an app developer point of view.

I think we can assume that the system HiDPI scaling factor is such that the system is UI is a legible size. Thus, the OS has to take care of both physical size and viewing distance via scale factor, and there's no reason for app developers to care about these variables directly, or do anything different from the rest of the system.

As far as I can tell this holds across all device types (including outliers like TVs and watches) and I don't think any of them are special in this regard. There are of course other differentiating factors that are relevant for app developers (e.g. remote-only input for TVs, or touch-only for phones), but I don't think physical display size is among them.

I'd be happy to be proven wrong if someone has a counterexample, but so far I haven't seen any.

docs/xml/metainfo-component.xml
```diff
@@ -834,6 +834,78 @@
...
+				</para>
+				<para>
+					Relations of the display size can be defined using a <literal>compare</literal> property as described for <literal>requires</literal>/<literal>recommends</literal>.
+					If this property is not present, a value of <code>ge</code> (greater-or-equal) is implicitly assumed.
+				</para>
+				<para>
+					The <literal>display_width</literal> and <literal>display_height</literal> tags also accept one of the following text values,
+					which are implementation-defined and should roughly match the device class listed next to them:
+				</para>
+				<itemizedlist>
```
dcz-purism, line 854: I would avoid baking subjective measures into the stadard. I do understand that people want the simplest possible solution, but I'd stick to objective measures. Subjective definitions will change over time, causing bad user experience, e.g. "why are all applications too big for my 7 year old phone?" because the "medium" tablet is now better served by the "small" category, or "half of the applications don't fit" when someone makes a small phone which is 10cm in height but the software center accepts both "xsmall" and "small" because neither is really defined.

If anything, I would rather disseminate templates for those sizes, made up of objective measures, for copy and pasting by packagers.

bertob: I think I agree, seeing as things like minimum size need to be precise values that developers actually test against. I don't think it makes sense for apps to subscribe to something that might change from under them, because the real problem is making apps work at new sizes, not the definition of "small" changing over time.

agx:
> As far as I can tell this holds across all device types (including outliers like TVs and watches) and I don't think any of them are special in this regard. There are of course other differentiating factors that are relevant for app developers (e.g. remote-only input for TVs, or touch-only for phones), but I don't think physical display size is among them.

Display size and distance+distance can matter if you want to know if e.g. the thumb can reach the whole screen and some other caes but i'm not trying to make a case for physical size here but the contrary.

bertob:
>  if e.g. the thumb can reach the whole screen

That's a good point. I suppose it's an input rather than a display issue though, and what you're interested in isn't just the size of the display but also the size of the entire device (and perhaps even thickness)? A 5" display can be ok to use with one hand if there are no bezels, but if there's a 1cm bezel at the top and bottom it might not be anymore.

Perhaps for this a slighly more subjective "single-handed usability" metric that takes various factors into account might be better?

That said: If we had this, how would this information be used by app developers?

dcz-purism:
I've been thinking along the same lines: reachability would be "purer" if it was its own mode of input. Two issues: would people actually bother declaring "my application is good with a thumb only" as opposed to just setting the screen size requirements? Also, it kind of forces the admission that granular choices are acceptable: "requires multitouch", "requires a scroll control", "can be used entirely with the left thumb on landscape", "can be used keyboard only", etc.

ximion:
An important thing to not is that **this information is not for applications**, nor is it *for* application developers. It is information *for* tools like KDE Discover, GNOME Software, Cockpit, Flatpak, etc. provided *by* applications for filtering components and informing users as well as making automated decisions.

@bertob 's argument for logical pixels makes sense to me, even though this can be flipped around as well: You can expect the device to make the texts and UI elements readable for its logical pixel size, but you can also expect the same when you request a physical length in mm. I made the display's physical size the default mode, because it fits the purpose of limiting apps to formfactors better than a logical pixel size, which is configuration dependent. But I do not feel very strongly about that.

By going for logical pixels, as soon as you reduce your resolution, your screen becomes "small" for the software center and it will therefore reveal apps that are only designed for smaller screens. While if you increase the resolution on a tiny screen, desktop apps will show up because the screen is now "larger" (even though all UI elements are tiny now). I think that's not a very likely scenario, in the majority of cases systems will be configured well, but the very fact that by using logical pixels you will check for a device+configuration rather than just checking for physical device limits is something I do not like that much.
This property is ideally supposed to check for device properties and not for configuration.

One important thing to remember is that `display_size` is only part of the puzzle. Applications can already make recommendations on which input methods they support for more than a year, so if your app recommends only a large TV screen it's very unlikely that it will also recommend touch input (but I think that is a fringe case anyway).

@agx I wouldn't throw viewing distance into the mix unless some actual usecases come up for that. Currently I think if the screen is large enough, a larger viewing distance can be assumed and that's "good enough". This property in AppStream is not supposed to give any precise measurement like a phone shell would do, neither to provide any information relevant for drawing the app's UI. Its purpose is solely to allow filtering apps by screen sizes they are designed for, so Libre Office will require a Tablet-Laptop-sized screen, phone apps will require a small screen and adaptive apps will not use this property at all and just run on any screen size (and they will only set the `control` tag to advertise touch input capabilities in addition to keyboard & mouse).

docs/xml/metainfo-component.xml
ximion: > things like minimum size need to be precise values that developers actually test against

I don't think so - the point of the placeholder values is to not give precise measurements, as software centers don't need this. Of course, apps need precise pixel dimensions, but the get that info from the compositor at runtime.
The point of placeholders is to just give app developers a possibility to say "this app roughly runs on large/small/tiny screens" and leave it up to the software center to actually judge whether the device's screen fits that model.
This should avoid situations where a device is just a mm/pixel too small to show a particular app, even though it would run fine. By giving absolute values (which some developers will absolutely do and have to do!) you are very strict about the sizes of displays that your app will be installable on.

@dcz-purism The device class of these sizes shouldn't change - `xsmall` will always be "smaller then phone", any "medium" size will always be around a 13'' laptop. If we need bigger sizes in future, we would have to add new named placeholders (tiny? xxlarge?). That definitely would need clarification in the text if we keep placeholders in the spec, so software centers don't use completely arbitrary values and redefine "medium" to suddenly mean "medium-sized phones".
In the long run I expect this to sort itself out - as apps like Blender add a "larger equal medium display size" requirement, software centers will not redefine "medium" to mean "medium phones" as otherwise Blender would show up on the tiny screen.
This certainly needs clarification though.

agx:
@ximion i only brought up viewing distance since for e.g. projectors the physical `display-size` is ill defined - but as i said: i wouldn't even worry about that for now since from what i've read here so far the only things that software centers (and also shells) would care about atm is logical pixels (and maybe input types).

docs/xml/metainfo-component.xml
dcz-purism: This causes two problems: sizes are on a continuum, and it's easily predictable that intermediate and extreme sizes will appear. Settling on a small number of values and waiting until things break to add easily predictable missing ones doesn't sound like a good idea to me.

The other issue that subjective measures are as much defined as fiat, as by usage. What recourse do we have when copy-pasting the wrong value becomes commonplace? I'm not convinced that asking nicely is going to prevent drift either.

bertob:
> I made the display's physical size the default mode, because it fits the purpose of limiting apps to formfactors better than a logical pixel size, which is configuration dependent. But I do not feel very strongly about that.

The problem is that it doesn't guarantee anything concrete with regard to whether an app will actually work on your display in the current configuration, which is the entire point of filtering apps in app stores, making it kind of useless for our purposes. The exact number of logical pixels an app scales down to is easily measured, but I really don't think we want app developers physically measuring their phones and putting that in metadata...

If you increase/decrease the resolution/scale factor on a device that impacts which apps fit the screen, so it makes sense that this changes what's available in Software. I get your point about device properties vs. configuration, but in practice the scale factor is basically part of device properties here, because there's generally only one good option (e.g. 2x on the Librem 5) and the device is unusable if you misconfigure it.

The idea with adaptive design is precisely that it doesn't matter whether you're running an app on a phone, desktop, TV, etc. as long as it fits the screen and supports the relevant input methods. As far as I can tell bringing physical sizes into this doesn't give us anything for displays (though it might be relevant for input).

docs/xml/metainfo-component.xml
ximion: Android does exactly this, allowing no absolute dimensions to be set - so I am fairly confident that this will work just as well for us.
Placeholder sizes would only be used by developers if they really do not care about exact dimensions, once they do they wouldn't make the change. Also, device vendors are expected to do sensible things, and once an established ecosystem exists, I am sure they will.

On the copy & paste matter: That has been a massive issue since basically forever, and is one of the reasons why the MetaInfo Creator exists: https://www.freedesktop.org/software/appstream/metainfocreator/#/
But in general, if people copy&paste the wrong values, it's their fault and a bug that they should fix. The AppStream lint tool will be able to find & fix any obvious cases, but for the rest, the developer has to watch out for it.

ximion:
> The problem is that it doesn't guarantee anything concrete with regard to whether an app will actually work on your display in the current configuration, which is the entire point of filtering apps in app stores [...]

That depends on what you actually really want to test for - the device, or the configuration.

> The idea with adaptive design is precisely that it doesn't matter whether you're running an app on a phone, desktop, TV, etc. as long as it fits the screen and supports the relevant input methods. As far as I can tell bringing physical sizes into this doesn't give us anything for displays (though it might be relevant for input).

Yes - fortunately the input part is already handled by a different property, so we don't need to worry about it :-)

ximion:
Here is a new revision with (most?) feedback incorporated. I originally wanted to defer adding the mm unit type, but given that both Qt and GTK+ have easy API for that and it's simple to implement and test, we can just add it now.
The two with/height tags are also now just one, checking for the shortest display side by default.

docs/xml/metainfo-component.xml
bertob: > This should avoid situations where a device is just a mm/pixel too small to show a particular app, even though it would run fine.

Except that's not true. If the app needs one pixel more than the screen width, it does not run fine, and shouldn't be shown in software centers. I really don't think introducing these shorthands buys us anything from an app developer or software center point of view, but it has the potential to muddy the waters so we'll end up with unreliable metadata.

docs/xml/metainfo-component.xml
```diff
@@ -834,6 +834,89 @@
 				</listitem>
 			</varlistentry>
 
+			<varlistentry id="tag-requires-recommends-display-width-height">
+				<term>&lt;display_length/&gt;</term>
+				<listitem>
+				<para>
+					Set a relation to the display length defined as an integer value in either <emphasis>millimeters</emphasis> or <emphasis>logical pixels</emphasis>
```
bertob, line 841: I'm still not sure what the potential use cases for millimeters are, so unless there's something else I'm missing I'd stick to having logical pixels only.

ximion: It's mimicking the existing Qt and GTK+ APIs as well as future-proofing the design (some adaptive apps may care more about the form factor than the drawing distance).

ximion:
If there's no further comment, I would move forward to implementing the current draft, only removing the `side` property for now (as we can always add that if we need it, while any other changes are harder to make).
Maybe some new issues will come up while implementing this, but if things go smoothly this feature may be in the (hopefully very soon to be released) AppStream 0.12.12 release.

docs/xml/metainfo-component.xml
Exalm: Which GTK APIs? GdkMonitor etc? Not sure what an app would use them for tbh, other than misusing them like in https://github.com/stdmedoth/simplenav/blob/dcfe229abcb1a6e31321407ae8855b8ad4c74fc7/src/init.c#L18-L20

ximion:
I've reduced the scope of the initial `display_length` tag a bit (only a `side` property allowed, only `px` as unit type for now) and added some more clarification text to the specification.
Also, I implemented the current specification draft. I still don't like it 100%, but I also have no ideas how to do this differently and achieve something that's better and still versatile enough.
I would merge this code soon to be part of the upcoming AppStream 0.12.12 release (maybe released next week).

Please note that the `display_length` spec is relies *very* heavily on the assumption that it is the shortest screen edge that apps care most about. It also, by doing that, kind of implicitly assumes that screens can be rotated, which is a bit mobile-centric. But since anything that isn't running on small or very large devices will have a `<display_length>medium</display_length>` recommendation anyway or not use this tag at all, it's probably fine.

PR #287, opened 2020-11-12
Set FD_CLOEXEC on the LMDB FD manually

Vogtinator:
Currently the FD referring to the appstreacm-cache-FOO.mdb is leaked into
child processes. The only way to fix is in a race-free way is by passing
O_CLOEXEC when opening it inside LMDB, but that's currently not done.

### DIFF
--- a/src/as-cache.c
+++ b/src/as-cache.c
@@ -35,6 +35,8 @@
 
 #include <lmdb.h>
 #include <errno.h>
+#include <fcntl.h>
+#include <unistd.h>
 #include <glib/gstdio.h>
 
 #include "as-utils-private.h"
@@ -723,6 +725,7 @@ as_cache_open (AsCache *cache, const gchar *fname, const gchar *locale, GError *
 	gboolean nosync;
 	gboolean readonly;
 	g_autoptr(GMutexLocker) locker = NULL;
+	int db_fd;
 
 	/* close cache in case it was open */
 	as_cache_close (cache);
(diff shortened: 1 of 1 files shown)


src/as-cache.c
```diff
@@ -829,6 +832,16 @@ as_cache_open (AsCache *cache, const gchar *fname, const gchar *locale, GError *
...
 
+	/* set FD_CLOEXEC manually. LMDB should do that, but it doesn't:
+	   https://www.openldap.org/lists/openldap-bugs/201702/msg00003.html */
+	rc = mdb_env_get_fd (priv->db_env, &db_fd);
+	if (rc == MDB_SUCCESS) {
+		db_fd_flags = fcntl (db_fd, F_GETFD);
+	}
+	if (db_fd_flags != -1) {
+		fcntl (db_fd, F_SETFD, db_fd_flags | FD_CLOEXEC);
+	}
```
ximion, line 843: Could you remove the extra brackets? If there's just a single line following an if statement, they should be omitted in AppStream's coding style

src/as-cache.c
```diff
@@ -829,6 +832,16 @@ as_cache_open (AsCache *cache, const gchar *fname, const gchar *locale, GError *
 		goto fail;
 	}
 
+	/* set FD_CLOEXEC manually. LMDB should do that, but it doesn't:
+	   https://www.openldap.org/lists/openldap-bugs/201702/msg00003.html */
+	rc = mdb_env_get_fd (priv->db_env, &db_fd);
+	if (rc == MDB_SUCCESS) {
+		db_fd_flags = fcntl (db_fd, F_GETFD);
```
ximion, line 839: Could you move the `db_fd_flags` variable into this block, so it only is visible when we actually need it, and also move the `if (db_fd_flags != -1)` into this block as well, to logically group this together?

ximion (changes requested):

ximion:
Oof, I kind of implicitly expected LMDB to do the right thing here and already use `O_CLOEXEC` - good catch! The race condition isn't great, but I think something we can live with until LMDB gets changed upstream.
Thank you for the patch!

src/as-cache.c
Vogtinator: It looked to me like it was using the old C style to declare all variables at the function start...

Done.

ximion: Yeah, some older parts of the code do, but for new code we can use fancy "new" things! - Even the `cleanup` attribute and `g_autofree`, which is one of the best things that ever happened to a C developer ;-)

ximion:
Looks good to me, thank you for the quick changes :-)

Vogtinator:
Thanks for the quick merge!

PR #290, opened 2020-11-30
Fix element ownership of returned PtrArrays

ricotz:


ximion:
Looks good to me, thank you a lot for the patch!

PR #298, opened 2021-01-12
Do not propagate a null GError

aleixpol:
I used to get:
```
** (process:12057): DEBUG: 05:26:52.319: No component found in desktop-entry file: SparkleShare.Autostart.desktop

(process:12057): GLib-CRITICAL **: 05:26:52.319: g_propagate_error: assertion 'src != NULL' failed
```

Possibly because the desktop file is broken, but it's good not to throw critical errors like that.


ximion:
Looks like this could use a test! - Apparently the testsuite only tests good desktop-entry files, not broken ones.
Thanks for the PR!

PR #302, opened 2021-02-10
build: Use c11 rather than gnu11 and define vendor extensions separately

pwithnall:
This makes the code a little more amenable to being built as a
subproject for projects like gnome-software, which use standard C.

See `man 7 feature_test_macros` for details of the vendor extension
macros.


ximion:
I would probably just have enabled `_GNU_SOURCE` to be done with it, but it's probably indeed better to be a bit more restrictive.

Issue #312, opened 2021-04-08
Clarify <project_license> for documentation


pwithnall:
We’ve come across [an issue](https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1200) recently where a project gave its license as
```
<project_license>GPL-2.0+ and CC-BY-SA-3.0</project_license>
```

I assume (but am [waiting to confirm](https://gitlab.gnome.org/GNOME/gnome-mahjongg/-/issues/28)) that this is intended to convey that the code is under `GPL-2.0+` and the documentation is under `CC-BY-SA-3.0`. However, in SPDX, `and` means the terms of both licenses must be complied with, which means that the code must also be licensed under `CC-BY-SA-3.0` and the documentation must also be licensed under `GPL-2.0+`.

`CC-BY-SA-3.0` is [not FSF or OSI approved](https://spdx.org/licenses/) so this has the end result of `as_license_is_free_license()` saying that this project is proprietary.

I *think* the fix here is for the project to simply use:
```
<project_license>GPL-2.0+</project_license>
```
since that’s the license for the project as a whole, and `CC-BY-SA-3.0` is only used for a portion of it.

That leads to the question of how to avoid this situation in future. My suggestion would be the following, but I’m sure many other solutions are possible and might be better:
 * Clarify in the [spec](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-project_license) that `project_license` is for the top-level license of the project (typically the one applied to the code, for a software project)
 * Potentially add a new `<documentation_license>` element for exposing the license of the project’s documentation independently of `<project_license>`

ximion:
Interesting! I agree with your issue description and solution suggestion - `project_license` is a SPDX license expression for the project-as-a-whole / the actual code, so `AND` is definitely not correct in this case.
That clarification should be added.
I am a bit wary about adding something like `documentation_license`, unless there's a clear use for it - for Debian, the `debian/copyright` file for a software package usually contains a full breakdown down to individual file level as to which license they are subjected to. I fear that we'd be replicating something like this eventually once we allow `documentation_license`, as more projects will then want to further clarify their licenses by adding more subcomponent-licenses to their metainfo files.
The other side of the argument is of course that documentation is usually the thing with a separate license from the actual code, so having that tag would cover most "different license exists" cases, while also being a bit more complete and honest.
Then the question remains whether a project with a non-free documentation license is also "non-free" or not...
[...]

PR #313, opened 2021-04-13
spec: Add a paragraph about which license to put in <project_license>

pwithnall:

Helps: #312


ximion:
Looks good to me! I kind of wonder whether we should highlight this even more by putting this information in a `<note>` box and mark it as hint or even important...
But having it in the text is a good start! Thank you for writing the change!

PR #318, opened 2021-05-27
as-content-rating: Fix a typo in the type macro for AsContentRating

pwithnall:
This is technically an API break, but since it was a violation of the
GObject naming conventions, it seems likely that nobody had encountered
it before (or relied on it), otherwise it would have been reported.

### DIFF
--- a/src/as-content-rating.c
+++ b/src/as-content-rating.c
@@ -1578,6 +1578,6 @@ AsContentRating*
 as_content_rating_new (void)
 {
 	AsContentRating *content_rating;
-	content_rating = g_object_new (AS_TYPE_CONTENT, NULL);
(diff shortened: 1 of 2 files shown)


pwithnall:
This is an API break. If you’d prefer to fix this without breaking API, I could put together an alternative patch which maintains a backwards-compatibility `#define`.

ximion:
Brr, such an annoying error! Looks like a string-replace mistake.
Since this doesn't break ABI and a quick search in open source code found no users of the old macro *and* it is easy to fix and will yield a backwards-compatible binary, I am fine with merging the patch.
The only annoying thing is that making code that builds with older and new libas versions will need a version check for this if it uses the macro...

Thanks for finding this and providing a patch!

PR #319, opened 2021-06-03
Fix example for YAML Launchable field

robert-ancell:


ximion:
Oh, something slipped by the great string replace years ago!
Thanks for noticing and fixing it! :-)

PR #320, opened 2021-06-03
component: Don't strip ";" from keywords before translating them

iainlane:
We're failing to translate strings now, when the keywords ends with a
";" as they almost always do, e.g.:

  Keywords=Drivers;Repositories;Repository;PPA;

because we strip the trailing semicolon off before translating, in order
to avoid having empty elements in the output.

Instead what we can do is process the list passed to
`as_component_set_keywords` and remove empties at that point. This
allows us to not modify the string retrieved from the desktop file, and
fixes finding translations.

### DIFF
--- a/src/as-component.c
+++ b/src/as-component.c
@@ -1301,14 +1301,25 @@ void
 as_component_set_keywords (AsComponent *cpt, gchar **value, const gchar *locale)
 {
 	AsComponentPrivate *priv = GET_PRIVATE (cpt);
+	g_autoptr(GPtrArray) keywords = NULL;
 
 	/* if no locale was specified, we assume the default locale */
 	if (locale == NULL)
 		locale = as_component_get_active_locale (cpt);
 
+	keywords = g_ptr_array_new ();
+
+	if (value != NULL) {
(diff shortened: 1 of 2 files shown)


iainlane:
I found this while I was trying to run asgen locally to reproduce https://github.com/ximion/appstream-generator/issues/92, no idea why this didn't happen all the time, new code?

ximion:
Just to make sure I understood the problem: AS drops the trailing semicolon first before the Ubuntu-specific l10n-hook is invoked, therefore the corresponding string won't be found in the translation file?
I don't know why we didn't see this yet... Potentially the respective code was there, but not used in asgen yet...
The patch itself looks good to me, pushing the sanitization code into `as_component_set_keywords` is a good idea as well.

iainlane:
You got it. As we're ending up passing this string (which came from the .desktop file) to gettext then we can't modify it at all before doing that, or it won't be found.

ximion:
A unit test would be nice for this, as I am breaking Ubuntu far too easily... But that's also a bigger thing to write (although with the new callback function, we could just mock all the mo-file reading and extraction...), so let's merge this bugfix first :-)

PR #321, opened 2021-06-04
compose: Check optipng is there before we use it

iainlane:
There are two places we can beef up the checks for optipng

  - When enabling its usage, make sure we found it first, and refuse to
    enable if we didn't.
  - In case `asc_compose_set_optipng_binary (NULL)` was called, check
    before spawning that we have *a* path. It's a programmer error to
    call `g_spawn_sync ()` with `NULL` in `argv[0]`, and the program
    will crash in that case. Also treat `NULL` as disabling optipng when
    passed to `asc_compose_set_optipng_binary`.

In case the binary is removed mid-run or an invalid binary is passed to
`asc_compose_set_optipng_binary`, the `g_spawn_sync` call itself will
fail and we'll report a proper GError. This is the best way to handle
it, as we avoid TOCTTOU races that way.

### DIFF
--- a/compose/asc-image.c
+++ b/compose/asc-image.c
@@ -204,6 +204,7 @@ asc_optimize_png (const gchar *fname, GError **error)
 {
 	gint exit_status;
 	gboolean r;
+	const gchar *optipng_path;
 	g_autofree gchar *opng_stdout = NULL;
 	g_autofree gchar *opng_stderr = NULL;
 	g_autofree const gchar **argv = NULL;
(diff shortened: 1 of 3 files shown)


compose/asc-image.c
```diff
@@ -212,8 +213,17 @@ asc_optimize_png (const gchar *fname, GError **error)
 	if (!asc_globals_get_use_optipng ())
 		return TRUE;
 
+	optipng_path = asc_globals_get_optipng_binary ();
+	if (optipng_path == NULL) {
```
iainlane, line 217: Given the checks in the setter(s), maybe this could be an assert instead

compose/asc-image.c
```diff
@@ -212,8 +213,17 @@ asc_optimize_png (const gchar *fname, GError **error)
...
+		g_set_error (error,
+			     ASC_IMAGE_ERROR,
+			     ASC_IMAGE_ERROR_FAILED,
+			     "optipng not found on $PATH");
```
ximion, line 221: Nitpick: "in PATH" is probably better...

tests/test-compose.c
```diff
@@ -493,6 +498,32 @@ test_compose_desktop_entry ()
...
+	g_setenv ("PATH", fixture->path, TRUE);
+	g_clear_pointer (&fixture->path, g_free);
+}
+
+static void
+test_compose_optipng_not_found (Fixture *fixture, gconstpointer user_data)
+{
+	g_test_expect_message (G_LOG_DOMAIN,
+			       G_LOG_LEVEL_CRITICAL,
+			       "*Refusing to enable optipng: not found in $PATH");
```
ximion, line 521: Is the asterisk prefix intended?

compose/asc-image.c
iainlane: yeah, typo I guess, that's what it is in the other places where I used a similar string

PR #322, opened 2021-06-07
utils: Don't strip modifiers when stripping encoding

iainlane:
We have some locales which are dupes if we do this, for example ca_ES.UTF-8 and ca_ES.UTF-8@valencia. The latter of these should become ca_ES@valencia in the output. That is what `locale -a` shows.
    
Previously `as_locale_strip_encoding ()` modified the passed-in string in place. However, in the one place where don't `g_strdup ()` the string before passing to this function, it is the key in a `GHashTable`. We can't do this, and only get away with it because the hash table isn't touched after this call. Fix the function to instead return a newly allocated string, and drop the `g_strdup` calls from the other call sites.
    
Add a small test for this function too.
    
https://github.com/ximion/appstream-generator/issues/92

### DIFF
--- a/src/as-utils.c
+++ b/src/as-utils.c
@@ -776,16 +776,12 @@ as_is_cruft_locale (const gchar *locale)
  * as_locale_strip_encoding:
  *
  * Remove the encoding from a locale string.
- * The function modifies the string directly.
+ * The function returns a newly allocated string.
  */
 gchar*
-as_locale_strip_encoding (gchar *locale)
+as_locale_strip_encoding (const gchar *locale)
 {
-	gchar *tmp;
-	tmp = g_strstr_len (locale, -1, ".UTF-8");
-	if (tmp != NULL)
-		*tmp = '\0';
-	return locale;
+	return as_str_replace (locale, ".UTF-8", "");
 }
 
 /**
--- a/src/as-utils.c
+++ b/src/as-utils.c
@@ -593,66 +593,55 @@ as_ptr_array_to_strv (GPtrArray *array)
 }
 
 /**
- * as_gstring_replace:
- * @string: The #GString to operate on
- * @search: The text to search for
- * @replace: The text to use for substitutions
+ * g_string_replace:
+ * @string: a #GString
+ * @find: the string to find in @string
+ * @replace: the string to insert in place of @find
+ * @limit: the maximum instances of @find to replace with @replace, or `0` for
+ * no limit
  *
- * Performs multiple search and replace operations on the given string.
+ * Replaces the string @find with the string @replace in a #GString up to
+ * @limit times. If the number of instances of @find in the #GString is
+ * less than @limit, all instances are replaced. If @limit is `0`,
+ * all instances of @find are replaced.
  *
- * Returns: the number of replacements done, or 0 if @search is not found.
+ * Returns: the number of find and replace operations performed.
  **/
 guint
-as_gstring_replace (GString *string, const gchar *search, const gchar *replace)
+as_gstring_replace (GString *string, const gchar *find, const gchar *replace, guint limit)
 {
-	gchar *tmp;
-	guint count = 0;
-	gsize search_idx = 0;
-	gsize replace_len;
-	gsize search_len;
+#if GLIB_CHECK_VERSION(2,68,0)
+	return g_string_replace (string, find, replace, limit);
+#else
+	/* note: This is a direct copy from GLib upstream (with whitespace
+	 * fixed spaces to tabs and with style fixed). Once we can depend on
+	 * GLib 2.68, this copy should be dropped and g_string_replace() used
+	 * instead.
+	 *
+	 * GLib is licensed under the LGPL-2.1+.
+	 */
+	gsize f_len, r_len, pos;
+	gchar *cur, *next;
+	guint n = 0;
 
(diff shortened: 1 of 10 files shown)


iainlane:
btw, bonus change to replace `as_gstring_replace` with upstream's new version, but I can drop that commit if you don't want it

src/as-utils.c
```diff
@@ -603,58 +603,33 @@ as_ptr_array_to_strv (GPtrArray *array)
...
+		g_string_insert (string, pos, replace);
+		cur = string->str + pos + r_len;
+		n++;
+	}
 
-	return count;
+	return n;
 }
 
+
```
robert-ancell, line 632: Nit: added an extra newline here.

src/as-utils.c
```diff
@@ -603,58 +603,33 @@ as_ptr_array_to_strv (GPtrArray *array)
  * Returns: the number of replacements done, or 0 if @search is not found.
```
robert-ancell, line 603: `@search` > `@find`

src/as-utils.c
```diff
@@ -603,58 +603,33 @@ as_ptr_array_to_strv (GPtrArray *array)
...
-		}
-		search_idx += replace_len;
-		count++;
-	} while (TRUE);
+	f_len = strlen (find);
+	r_len = strlen (replace);
+	cur = string->str;
+
+	while ((next = strstr (cur, find)) != NULL)
+	{
```
robert-ancell, line 621: Brace should be on the end of the previous line to match existing style.

robert-ancell:
Other than the minor changes above, LGTM, thanks @iainlane!

src/as-utils.c
iainlane: I've not fixed style like that (only spaces to tabs) as this is a direct copy-paste from glib.

iainlane: (added a comment to make that clear. I can update the style to match AS's if that's desired)

iainlane:
Merci for the review!

src/as-utils.c
ximion: Yes, please make it follow the style of the rest of the AppStream code :-)

src/as-utils.c
```diff
@@ -595,64 +595,44 @@ as_ptr_array_to_strv (GPtrArray *array)
...
-	gsize search_idx = 0;
-	gsize replace_len;
-	gsize search_len;
+as_gstring_replace (GString *string, const gchar *find, const gchar *replace)
+{
+	/* note: This is a direct copy from GLib upstream (with whitespace
+	 * fixed spaces to tabs). Once we can depend on GLib 2.68, this copy
+	 * should be dropped and g_string_replace() used instead.
+	 *
+	 * GLib is licensed under the LGPL-2.1+.
```
ximion, line 612: Can you throw the whole code of this function behind a GLib version check macro, so we fall back to the GLib-internal version as soon as it is available with newer GLib? That also makes the code self-explanatory and is easy to grep for once we do raise the minimum GLib version.
See `as_date_time_format_iso8601` for one such example: https://github.com/ximion/appstream/blob/master/src/as-utils.c#L1744-L1782 (there's quite a few more in AppStream's code)

src/as-yaml.c
```diff
@@ -641,6 +641,7 @@ as_yaml_emit_sequence_from_str_array (yaml_emitter_t *emitter, const gchar *key,
 static void
 as_yaml_localized_list_helper (gchar *key, gchar **strv, yaml_emitter_t *emitter)
 {
+	g_autofree gchar *locale_noenc;
```
ximion, line 644: `g_autofree gchar *locale_noenc = NULL;`, otherwise we'll crash if `strv == NULL`.

ximion (changes requested):
Looks good except for the mentioned changes. Since `as_gstring_replace` is no public API, can you, besides conditionally redirecting it to `g_string_replace` if that's available, also add the `limit` parameter and set that to `1` when using it to replace the UTF-8 "sometimes-suffix"? It'll probably not actually improve performance by an amount that matters, given how short the string is, but it's likely still nice to have.

iainlane:
All done. Please check the SPDX changes, as I don't know what those input strings are to know if it should be `1` or `0`. (Might be easier for you to fix that up directly in this branch if you want - you should be able to force push to it and then merge.)

src/as-utils.c
```diff
@@ -593,85 +593,76 @@ as_ptr_array_to_strv (GPtrArray *array)
...
 /**
- * as_gstring_replace:
- * @string: The #GString to operate on
- * @search: The text to search for
- * @replace: The text to use for substitutions
- *
- * Performs multiple search and replace operations on the given string.
- *
- * Returns: the number of replacements done, or 0 if @search is not found.
+ * g_string_replace:
```
iainlane, line 596: that's a typo (c+p from glib)

src/as-utils.h
```diff
@@ -118,7 +118,8 @@ guint			as_utils_data_id_hash (const gchar *data_id);
 
 guint			as_gstring_replace (GString *string,
 					    const gchar *search,
-					    const gchar *replace);
+					    const gchar *replace,
+					    guint limit);
```
iainlane, line 122: seems like an API & ABI break; maybe we can't implement this suggestion?

ximion (approved):

Issue #326, opened 2021-06-10
YAML firmware code doesn't match documentation.


robert-ancell:
In the [firmware documentation](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Firmware.html#tag-firmware-provides) it states that `runtime` firmware has a file, and `flashed` firmware has a GUID.

In `as_component_yaml_emit_provides()` the following code seems to have this reversed:
```C
if (fw_runtime != NULL) {
        for (i = 0; i < fw_runtime->len; i++) {
                const gchar *value = g_ptr_array_index (fw_runtime, i);
                as_yaml_mapping_start (emitter);
                        as_yaml_emit_entry (emitter, "type", "runtime");
                as_yaml_emit_entry (emitter, "guid", value);
                        as_yaml_mapping_end (emitter);
        }       
}

if (fw_flashed != NULL) {
        for (i = 0; i < fw_flashed->len; i++) {
                const gchar *value = g_ptr_array_index (fw_flashed, i);
                as_yaml_mapping_start (emitter);
                        as_yaml_emit_entry (emitter, "type", "flashed");
                as_yaml_emit_entry (emitter, "file", value);
                        as_yaml_mapping_end (emitter);
        }
}
```

Is the YAML code wrong?

robert-ancell:
@hughsie probably knows this best.

hughsie:
The docs are correct and the code is wrong.

robert-ancell:
Thanks @hughsie, should be fixed in #328.

ximion:
Good catch!
The thing is, even though I don't want it to be this way, the YAML variant of AppStream is very much a second-class citizen and underdocumented and also lacking a few tests & features.
It primarily exists because early on, the Debian FTP masters refused to have XML in the archive but did accept YAML, and since programs were to consume AppStream data exclusively through an AppStream database according to the old projects's drafts or the libappstream library respectively, I did think just implementing the YAML variant was the way of least resistance to improve AppStream adoption - then appstream-glib came along and we dropped the database, and things got a lot more complicated and much larger than I ever thought ^^

PR #328, opened 2021-06-10
Fix YAML having wrong names for the firmware data

robert-ancell:
Fixes https://github.com/ximion/appstream/issues/326


ximion:
Excellent work, thank you for the patch!

PR #329, opened 2021-06-10
Fix flashed firmware generating incorrect XML

robert-ancell:


ximion:
Wow, what an oversight... I am finde with merging this, but could you maybe also write a test case for this, or just add this to one of the existing tests? If you don't have the time, I'll merge this later and then add a test myself.

robert-ancell:
Tests updated.

ximion:
Thanks! :-)

PR #332, opened 2021-07-05
qt: Make Pool's parent class public

aleixpol:
It's not the default but what we generally do. It allows us to call
parent class methods normally.


ximion:
I think the base class not being public was a mistake here, actually. Thanks for the patch!

PR #347, opened 2021-08-29
ci: Add Fedora to the CI environment

Conan-Kudo:
As Fedora ships updated GNOME components often earlier than anyone else,
it makes sense to test this to ensure that this continues to work.

### DIFF
--- a/tests/ci/Dockerfile-fedora-latest
+++ b/tests/ci/Dockerfile-fedora-latest
@@ -1,0 +1,14 @@
+#
+# Docker file for AppStream CI tests on Fedora targeting latest stable
+#
+FROM registry.fedoraproject.org/fedora:latest
+
+# prepare
+RUN mkdir -p /build/ci/
+
+# install build dependencies
+COPY install-deps-rpm.sh /build/ci/
+RUN chmod +x /build/ci/install-deps-rpm.sh && /build/ci/install-deps-rpm.sh
+
+# finish
+WORKDIR /build
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -3,6 +3,42 @@ name: Build Test
 on: [push, pull_request]
 
 jobs:
+  build-fedora-latest:
+    name: Fedora Latest Stable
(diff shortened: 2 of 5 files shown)


ximion:
Do we really need a sanitized build on Fedora as well? I've deliberately put it only on one OS because it's quite slow (and getting faster feedback from the CI is often useful).
Also, Fedora Noob question: How often does rawhide break due to e.g. dependency issues in unrelated packages? If it's rarely/never, then we can leave everything as-is, otherwise we may want to ignore failures for that configuration. I'd also be fine with just testing it out, if there's no definitive answer ^^
Otherwise the PR looks fine, thanks for creating it! :-)

tests/ci/Dockerfile-fedora-latest
```diff
@@ -0,0 +1,14 @@
+#
+# Docker file for AppStream CI tests on Fedora Rawhide
```
ximion, line 2: Fedora Latest ;-)

Conan-Kudo:
> Do we really need a sanitized build on Fedora as well? I've deliberately put it only on one OS because it's quite slow (and getting faster feedback from the CI is often useful).

Fedora Rawhide has the literally newest GCC, which I figured would be quite valuable for this.

> Also, Fedora Noob question: How often does rawhide break due to e.g. dependency issues in unrelated packages? If it's rarely/never, then we can leave everything as-is, otherwise we may want to ignore failures for that configuration. I'd also be fine with just testing it out, if there's no definitive answer ^^

It's relatively rare these days. The main breakage is around Python rebases usually, but we've streamlined that to avoid that problem as much as possible.

> Otherwise the PR looks fine, thanks for creating it! :-)

👍🏾

tests/ci/Dockerfile-fedora-latest
Conan-Kudo: Fixed!

ximion:
Latest GCC probably includes the dev release sometimes... Which might not be the worst thing to test against, so let's give this a try!

Issue #345, opened 2021-08-11
Spec: Define whether <languages> should list en_US explicitly


pwithnall:
Almost all applications (which use gettext, at least) are written with their C (untranslated) locale in `en_US`. This means that they never have an explicit `en_US` translation, but are always 100% translated to `en_US`.

Should the [`<languages>`](https://www.freedesktop.org/software/appstream/docs/chap-CollectionData.html#tag-ct-languages) element list `en_US` explicitly, in which case it should probably be injected by the appstream XML generator?

Or should the spec say that the application’s C locale is never listed as a `<lang>` element, and should be assumed to be `en_US`?

If we go with the second option, perhaps the assumption about the C locale being `en_US` could be overrideable (in the few cases where an application is written in something other than US English by default) by setting an attribute like `<languages default-language="es_ES">` and defining that language as 100% translated.

I can put together a merge request to update the specification either way, but need to know what approach would be preferable :slightly_smiling_face:

hughsie:
My assumption was always that the default language was C == en_US.

pwithnall:
Yeah, and that’s a good assumption, but I can’t find anything in gettext which actually requires C == en_US. I might be missing it, or it might be the case that there’s a handful of apps out there where C is French or German or whatever.

ximion:
I actually came across a project where C was German, and two where C was Mandarin and Russian (which was horrendous for collaboration...), so I think the AppStream spec shouldn't blindly encode C == en_US.
I wonder if we can do some heuristics here... Otherwise having `<languages default="es_ES">` looks like a good idea, and also follows AppStream's "explicit is better than implicit" rule (Is `default` a good name? Maybe `template` or `source` is better?). If that property isn't there, we'll assume `en_US` as default, and libappstream could even synthesize a 100% translation entry for that language in those cases, which would solve this issue and be fully backwards compatible.
I guess for 99.9% of all projects, assuming en_US as template language is the right thing to do.

pwithnall:
Thanks for your thoughts @ximion! I’ll try and put together a draft merge request for this (and my other recent issues) in September, once I have a bit of headspace from the GNOME release.

pwithnall:
I’ve put some proposed spec changes together at https://github.com/ximion/appstream/pull/351 for discussion. I don’t think they should be merged until libappstream and appstream-compose have been changed to support the new functionality, but I didn’t want to work on that until the spec changes are largely agreed.

PR #351, opened 2021-09-08
spec: Add support for source locales

pwithnall:
This allows specifying the locale of the source translatable strings
in a component. The vast majority of the time, this is `en_US`, but
that’s not guaranteed. Add a `source-locale` attribute to allow changing
this default.

Helps: #345

### DIFF
--- a/src/as-translation.c
+++ b/src/as-translation.c
@@ -42,6 +42,7 @@ typedef struct
 {
 	AsTranslationKind	kind;
 	GRefString		*id;
+	GRefString		*source_locale;
 } AsTranslationPrivate;
 
 G_DEFINE_TYPE_WITH_PRIVATE (AsTranslation, as_translation, G_TYPE_OBJECT)
@@ -103,6 +104,7 @@ as_translation_finalize (GObject *object)
 	AsTranslationPrivate *priv = GET_PRIVATE (tr);
 
 	as_ref_string_release (priv->id);
+	as_ref_string_release (priv->source_locale);
 
 	G_OBJECT_CLASS (as_translation_parent_class)->finalize (object);
 }
@@ -174,6 +176,42 @@ as_translation_set_id (AsTranslation *tr, const gchar *id)
 }
 
 /**
+ * as_translation_get_source_locale:
+ * @tr: a #AsTranslation instance.
+ *
+ * The locale of the source strings for this component. If this has not been
+ * explicitly specified, `en_US` will be returned.
+ *
+ * Returns: (not nullable): The locale of the source strings for this component.
+ * Since: 0.14.6
+ */
+const gchar*
+as_translation_get_source_locale (AsTranslation *tr)
+{
+	AsTranslationPrivate *priv = GET_PRIVATE (tr);
+	return (priv->source_locale != NULL) ? priv->source_locale : "en_US";
+}
+
+/**
+ * as_translation_set_source_locale:
+ * @tr: a #AsTranslation instance.
+ * @locale: (nullable): The locale that the source strings are in, or %NULL if
+ *    unknown or default.
+ *
+ * Set the locale of the source strings for this component. In gettext, this is
+ * referred to as the `C` locale. It’s almost always `en_US`, but for some
+ * components it may not be.
+ *
+ * Since: 0.14.6
+ */
+void
+as_translation_set_source_locale (AsTranslation *tr, const gchar *locale)
+{
+	AsTranslationPrivate *priv = GET_PRIVATE (tr);
+	as_ref_string_assign_safe (&priv->source_locale, locale);
+}
+
+/**
  * as_translation_load_from_xml:
  * @tr: a #AsTranslation instance.
  * @ctx: the AppStream document context.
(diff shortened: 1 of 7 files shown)


docs/xml/collection-xmldata.xml
```diff
@@ -514,14 +514,23 @@
 						Each <code>&lt;lang/&gt;</code> node may have a <literal>percentage</literal> property, which describes the percentage value to which
 						a component has been translated.
 					</para>
+					<para>
+						The source locale for a component is assumed to be <code>en_US</code>. This is the language that the untranslated strings in the
+						component are written in. In GNU Gettext terminology, this is the <code>C</code> locale. If the component has a different source
+						language, it should be specified using the <code>source-locale</code> attribute on the <code>&lt;languages/&gt;</code> element.
+					</para>
+					<oara>
```
ximion, line 522: Typo, makes the CI complain about this PR currently ;-)

docs/xml/collection-xmldata.xml
```diff
@@ -514,14 +514,23 @@
 						Each <code>&lt;lang/&gt;</code> node may have a <literal>percentage</literal> property, which describes the percentage value to which
 						a component has been translated.
 					</para>
+					<para>
+						The source locale for a component is assumed to be <code>en_US</code>. This is the language that the untranslated strings in the
+						component are written in. In GNU Gettext terminology, this is the <code>C</code> locale. If the component has a different source
+						language, it should be specified using the <code>source-locale</code> attribute on the <code>&lt;languages/&gt;</code> element.
```
ximion, line 520: We should use snake_case instead of kebap-case for the new property, so `source_locale` please. This is simply because the rest of the specification uses snake_case for properties and tag names, and consistency is important :-)

ximion (changes requested):

ximion:
With the tag name style change and the para typo fixed, I think this is fine.
I do wonder though whether we maybe only need `source_locale` on the `translation` tag, and, since the `language` tags are usually auto-generated, could then make the generator code just "do the right thing" and insert a 100% en_US block or whatever source_locale was set to.
Is there any value in knowing the original language in a language tag block?

If we would go with the `source_locale` property on both `translation` and `languages` approach, I wonder if implementation-wise, libappstream should just abstract this away and "do the right thing", that is, add a "100% en_US" or "100% de_DE" entry if `source_locale` is either of the two. Then, when serializing the XML, it could restore the original source_locale property and drop the extra language value.
Would probably be a bit annoying to implement, but would save all clients from having to deal with this explicitly.

docs/xml/collection-xmldata.xml
pwithnall: Done

docs/xml/collection-xmldata.xml
pwithnall: Oops

pwithnall:
> With the tag name style change and the para typo fixed, I think this is fine.
> I do wonder though whether we maybe only need `source_locale` on the `translation` tag, and, since the `language` tags are usually auto-generated, could then make the generator code just "do the right thing" and insert a 100% en_US block or whatever source_locale was set to.

Sure, I’ve changed the spec wording and changes to go with that.

I’ll start working on the appstream-compose changes for this now, so please shout if something still looks wrong in the spec!

pwithnall:
> I’ll start working on the appstream-compose changes for this now, so please shout if something still looks wrong in the spec!

Done!

src/as-translation.c
```diff
@@ -194,6 +232,8 @@ as_translation_load_from_xml (AsTranslation *tr, AsContext *ctx, xmlNode *node,
 	if (priv->kind == AS_TRANSLATION_KIND_UNKNOWN)
 		return FALSE;
 
+	as_translation_set_source_locale (tr, (gchar*) xmlGetProp (node, (xmlChar*) "source_locale"));
```
ximion, line 235: Please use `as_xml_get_prop_value (node, "source_locale")` for new code (just a convenience macro, but it improves readability and makes refactoring easier).
You are also leaking memory here, as the result of `xmlGetProp` needs to be freed.

You can solve all of this with just a few convenience functions:
```c
as_ref_string_assign_transfer (&priv->source_locale,
                               as_xml_get_prop_value_refstr (node, "source_locale"));
```

src/as-translation.c
```diff
@@ -221,6 +261,9 @@ as_translation_to_xml_node (AsTranslation *tr, AsContext *ctx, xmlNode *root)
 	n = xmlNewTextChild (root, NULL, (xmlChar*) "translation", (xmlChar*) priv->id);
 	xmlNewProp (n, (xmlChar*) "type",
 			(xmlChar*) as_translation_kind_to_string (priv->kind));
+
+	if (priv->source_locale != NULL)
+		xmlNewProp (n, (xmlChar*) "source_locale", (xmlChar*) priv->source_locale);
```
ximion, line 266: Please use `as_xml_add_text_prop` here, it's a bit nicer for newly written code. So these whole two lines (you can drop the null check) would just become a single line:
```c
as_xml_add_text_prop (n, "source_locale", priv->source_locale);
```

ximion (changes requested):

ximion:
Looks good, except for the memory leak ^^
One peculiar thing about libappstream's code is that a lot of it was written without convenience helpers for the XML stuff at first, because adding `if (a != NULL)` and `(xmlChar*)` everywhere isn't so bad... you would think, until you have done that a billion times :-P
So, new code should use those helpers, but reading through all the old code to refactor 100% of it to use the new helpers so far wasn't worth the effort. I admit that the current mix is odd for people touching the code initially though.

src/as-translation.c
pwithnall: Oops, thanks for the suggestion

src/as-translation.c
pwithnall: Nice

pwithnall:
> So, new code should use those helpers, but reading through all the old code to refactor 100% of it to use the new helpers so far wasn't worth the effort. I admit that the current mix is odd for people touching the code initially though.

Makes sense. Branch updated :)

ximion:
Looks good, thank you for working on this! :-)

PR #353, opened 2021-09-08
spec: Mention the already-supported scale attribute of <icon/>

pwithnall:
As I understand it, these changes just document the existing semantics
and don’t change anything.

### DIFF
--- a/docs/xml/collection-xmldata.xml
+++ b/docs/xml/collection-xmldata.xml
@@ -270,23 +270,34 @@
 						<literal>stock</literal> icons are loaded from stock. The icon name should never include any file-extension or path.
 					</para>
 					<para>
-						<literal>cached</literal> icons are loaded from the AppStream icon cache. The icon tag should contain the icon file name, including it's
+						<literal>cached</literal> icons are loaded from the AppStream icon cache. The icon tag should contain the icon file name, including its
 						extension. It must not contain a full or relative path to the icon file.
+						This icon type may have <literal>width</literal> and <literal>height</literal> properties.
+						If targeting a hi-DPI screen, this icon type may have a <literal>scale</literal> property.
 					</para>
 					<para>
 						<literal>local</literal> icons are reserved for AppStream data installed by local applications or via 3rd-party application installers.
 						They should specify a full file path.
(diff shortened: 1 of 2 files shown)


ximion:
Wow, looks like this was indeed forgotten from the spec! The rule is that the spec change is committed first, and the implementation follows that, apparently on the scale property, that process either didn't exist yet or was ignored.
Thanks for noticing and creating the PR!

docs/xml/collection-xmldata.xml
```diff
@@ -270,23 +270,28 @@
 						<literal>stock</literal> icons are loaded from stock. The icon name should never include any file-extension or path.
 					</para>
 					<para>
-						<literal>cached</literal> icons are loaded from the AppStream icon cache. The icon tag should contain the icon file name, including it's
+						<literal>cached</literal> icons are loaded from the AppStream icon cache. The icon tag should contain the icon file name, including its
 						extension. It must not contain a full or relative path to the icon file.
+						This icon type may have <literal>width</literal> and <literal>height</literal> properties.
+						If targeting a hi-DPI screen, this icon type may have a <literal>scale</literal> property.
```
ximion, line 276: Could we briefly explain the semantic of "scale" here, even if it's obvious?
So something like `If targeting a hi-DPI screen, this icon type may have a <literal>scale</literal> property with an integer value > 1 to denote the icon's scale factor, e.g. <code>scale="2"</code> for 2x scaling.`
We could also refer to the Freedesktop Icon Theme Specification here, but that would likely be a bit too verbose.

ximion (changes requested):

docs/xml/collection-xmldata.xml
pwithnall: Done

ximion:
Looks good, thank you!

I do notice though that the icons are still explained for "collection metadata", while nowadays we should describe stuff for metainfo files first and only have the collection document describe differences (so this is currently backwards). That's an artifact from when the metainfo spec didn't exist, as collection data was first, and should probably be changed.
But that's out of scope for this PR :-)

PR #356, opened 2021-09-13
spdx: Make CC-BY-3.0 and CC-BY-SA-3.0 free

Kekun:
Without this, GNOME Weather is considered to be proprietary by GNOME
Software.

### DIFF
--- a/data/spdx-free-license-ids.txt
+++ b/data/spdx-free-license-ids.txt
@@ -35,7 +35,11 @@ BitTorrent-1.1
 CAL-1.0
 CAL-1.0-Combined-Work-Exception
 CATOSL-1.1
+CC-BY-2.5
+CC-BY-3.0
 CC-BY-4.0
+CC-BY-SA-1.0
+CC-BY-SA-3.0
 CC-BY-SA-4.0
 CC0-1.0
 CDDL-1.0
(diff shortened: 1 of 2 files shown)


ximion:
This data is auto-generated from [SPDX](https://spdx.dev/) data, and the free-ness is derived from whether a license is considered "free" by OSI or the FSF.
According to the source data, this is not the case: https://github.com/spdx/license-list-data/blob/master/json/licenses.json#L2794
I don't want to hack around the auto-generated in AppStream, rather SPDX should add a merked that the license is open-source, or the FSF or OSI should just approve the license.
Given that the CC licenses are not for code, I actually wonder though whether the license for GNOME Weather is set correctly - `project_license` should only be for *code* and shouldn't include the documentation license (if the source code is really CC licensed, then that's correct of course).

pwithnall:
> Given that the CC licenses are not for code, I actually wonder though whether the license for GNOME Weather is set correctly - `project_license` should only be for _code_ and shouldn't include the documentation license (if the source code is really CC licensed, then that's correct of course).

This looks like https://github.com/ximion/appstream/issues/312 / https://tecnocode.co.uk/2021/07/05/dont-generally-put-cc-by-sa-in-appdata/

ximion:
I really don't think this is the solution - I will probably augment the free-license-list by a list of licenses that are considered DFSG-free, as outlined in https://github.com/ximion/appstream/issues/357#issuecomment-926245405, that should also resolve this issue and still not make AppStream a legal-ish authority on what is free software and what is not.

PR #364, opened 2021-11-25
Fix various typos and grammar issues

mwleeds:


cassidyjames (approved):

ximion:
Thank you!
Just out of curiosity, did you find those manually or did you use a tool for that?

mwleeds:
Just found them by reading through the spec

ximion:
Neat, thank you a lot for caring! When reading though this so many times, and especially as non-native speaker, you start to unconsciously ignore these.
(If you had used some kind of tool, I would actually have added it to the CI - DAPS has a spellchecker, but that's better in interactive mode as it sometimes tries to correct stuff that shouldn't be changed).

PR #369, opened 2021-12-09
Fix mention of console apps in web apps docs

mwleeds:


ximion:
Eww, copy-paste error! Thanks!

PR #371, opened 2021-12-28
docs: document media_baseurl property of components

pabloyoyoista:
This property is inserted[1] into XML collection metadata by
appstream-generator, but is undocumented.

[1]https://github.com/ximion/appstream-generator/blob/master/src/asgen/engine.d#L318


ximion:
Jup, this was an experimental feature for a while, that kind of became the norm for collection data.
Thanks for the patch!

PR #372, opened 2021-12-31
Check missing release props

fortysixandtwo:
`appstream-util` checks for missing timestamp property, so we can do the same.

What prompted me to check this, was having had a typo `data=...` instead of `date=...` which was caught by `appstream-util`, but not by `appstreamcli`.

An alternative solution could iterate the properties of the release tag and check for unknowns (if that is preferred).

I'm also not sure if the info severity is a good choice or if it should be bumped to warning

### DIFF
--- a/src/as-validator-issue-tag.h
+++ b/src/as-validator-issue-tag.h
@@ -691,6 +691,11 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("The release is missing the `version` property."),
 	},
 
+	{ "release-time-missing",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
+	  N_("The release is missing either the `date` (preferred) or the `timestamp` property."),
+	},
 	{ "artifact-type-invalid",
 	  AS_ISSUE_SEVERITY_ERROR,
 	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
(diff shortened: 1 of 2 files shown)


src/as-validator-issue-tag.h
```diff
@@ -685,6 +685,11 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("The value set as release type is invalid."),
 	},
 
+	{ "release-property-missing",
```
ximion, line 688: Please split this into `release-version-missing` and e.g. `release-time-missing`, in order to adjust the priorities individually and give a better explanation (especially for the "either `date` or `timestamp` case).

src/as-validator-issue-tag.h
```diff
@@ -685,6 +685,11 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
...
+	  AS_ISSUE_SEVERITY_INFO,
```
ximion, line 689: This will have at least `WARNING` severity, as it violates the specification and should make the validation fail. I would say that a missing version would also definitely be an `ERROR` severity, as it will actually make the release data unusable on most application (I think libappstream will even refuse to read such a release).
If you want to make both takes errors, that would be fine with me too though.

ximion (changes requested):
A few minor changes on the issue tags and severities, but otherwise this patch looks good to me. Odd that this check was missing in the first place, thanks for noticing and creating a patch!

fortysixandtwo:
v2: Split into `release-version-missing` and `release-time-missing` and bumped to error severity (thanks @ximion for the review)

ximion:
This looks good now, thank you for the patch! :-)

Issue #375, opened 2022-01-07
Some icons are wrongly rescaled


antonio-rojas:
Since commit c57909568f0ffa568da251102a0403fed936a43c, many icons are wrongly rescaled when indexing metadata, and they take only the first quarter of the canvas. This affects most gnome application icons, for instance the ones from https://archlinux.thaller.ws/extra/os/x86_64/evolution-3.42.2-1-x86_64.pkg.tar.zst

Before c57909568f0ffa568da251102a0403fed936a43c:
![pre](https://user-images.githubusercontent.com/11243355/148586330-e00fd369-8e2d-40a9-8f7c-37a3d9203744.png)

After c57909568f0ffa568da251102a0403fed936a43c:
![post](https://user-images.githubusercontent.com/11243355/148586346-57d106f8-d11c-43e9-89e2-45178e98408c.png)

librsvg is at 2.52.5

heftig:
I think you shouldn't `cairo_scale` the surface in the path that uses the viewport, as that already takes care of rescaling the svg.

PR #376, opened 2022-01-07
compose: Fix SVG scaling with librsvg 2.52

heftig:
We should only scale the canvas in the legacy path, since the viewport
already takes care of scaling.

Also, while we're here:

- We should `cairo_save` before scaling, so that on error the
  `cairo_restore` doesn't leave the canvas in the scaled state.
- The SVG size is only used in the legacy path.
- We don't inspect the error so no need to have `tmp_error`.

Fixes: https://github.com/ximion/appstream/issues/375


ximion:
Sorry for the delay, and thank you for providing a patch!
This change looks good (except for the weird commented-out code, but that was already there previously).

PR #378, opened 2022-01-17
qt: Fix crash when moving a Component

aleixpol:
We need to invalidate the one we are moving, otherwise we get a 2x
unref.

### DIFF
--- a/qt/component.cpp
+++ b/qt/component.cpp
@@ -201,6 +201,7 @@ Component& Component::operator=(const Component& other)
 Component::Component(Component&& other)
     : m_cpt(other.m_cpt)
 {
+    other.m_cpt = nullptr;
 }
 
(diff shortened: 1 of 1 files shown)


ximion:
CI failures are unrelated. Thank you for the patch!

ximion:
I may have merged this too quickly, as it looks like if you don't create a new component for the moved component (`other.m_cpt = as_component_new ()` as well), then the other component will try to unref a nullpointer, which will either be a hard error or at least give a nasty warning (or am I missing something here?).

aleixpol:
I checked and it does warn when it finds a null object to unref.

https://github.com/ximion/appstream/pull/379

ximion:
Jup, sorry - should have actually reviewed this more carefully!

PR #380, opened 2022-01-25
Fix translate URL

JakobDev:
While looking at the source, I saw that UrlTranslate is defined in Header as part of UrlKind but not set.

btw:
Is there any reason why it is named UrlTranslate and not UrlKindTranslate like the other?

### DIFF
--- a/qt/component.cpp
+++ b/qt/component.cpp
@@ -139,6 +139,9 @@ Component::UrlKind Component::stringToUrlKind(const QString& urlKindString) {
     if (urlKindString == QLatin1String("contact")) {
         return UrlKindContact;
     }
+    if (urlKindString == QLatin1String("translate")) {
+        return UrlTranslate;
+    }
     return UrlKindUnknown;
 }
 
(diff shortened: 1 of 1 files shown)


ximion:
Wow, something went massively wrong here... Good catch!

JakobDev:
Should I rename UrlTranslate to UrlKindTranslate? It never worked and there are no Issues about it, so I don't think it is used in any code.

ximion:
> Should I rename UrlTranslate to UrlKindTranslate? It never worked and there are no Issues about it, so I don't think it is used in any code.

No, we can't do that as that would still be an API break (the enum value was valid afterall and some client could have used it in a `Component::url` call). I changed the code to properly deprecated the mistyped one in https://github.com/ximion/appstream/commit/f336132884652238438b3cf430d2a06dae24d305, so this should work now :-) (and with the next API/ABI break we can then remove this enum value fully)
Thank you for noticing this issue and providing a patch!

Issue #384, opened 2022-02-26
`flatpak search` assertion failure since port to libappstream: g_atomic_pointer_get (value_location) == 0


smcv:
Relevant package versions:

```
libappstream4:amd64	0.15.2-2
libglib2.0-0:amd64	2.70.4-1
```

with flatpak git HEAD, on Debian unstable. (Note that flatpak 1.12.x used libappstream-glib, but git HEAD has switched to libappstream.)

Steps to reproduce: run the Flatpak test suite

Expected result: tests succeed

Actual result: several tests fail with this assertion failure in libappstream:

```
(flatpak search:260854): GLib-CRITICAL **: 14:05:04.766: g_once_init_leave: assertion 'g_atomic_pointer_get (value_location) == 0' failed
.../tests/test-repo.sh: line 109: 260854 Trace/breakpoint trap   (core dumped) ${FLATPAK} search Hello > search-results
```

Backtrace:

```
#0  g_logv
    (log_domain=0x7f5e3982300e "GLib", log_level=G_LOG_LEVEL_CRITICAL, format=<optimized out>, args=<optimized out>)
    at ../../../glib/gmessages.c:1417
#1  0x00007f5e397d678f in g_log (log_domain=<optimized out>, log_level=<optimized out>, format=<optimized out>)
    at ../../../glib/gmessages.c:1455
#2  0x00007f5e3a1af451 in as_component_search_matches
    (cpt=cpt@entry=0x55cc24af1200 [AsComponent], term=term@entry=0x7ffc50fcf003 "org.test.Hello")
    at ../src/as-component.c:2942
#3  0x000055cc22a57ddd in flatpak_builtin_search
    (argc=<optimized out>, argv=<optimized out>, cancellable=<optimized out>, error=<optimized out>)
    at /home/smcv/src/flatpak/app/flatpak-builtins-search.c:296
#4  0x000055cc22a2e0e6 in flatpak_run (argc=<optimized out>, argc@entry=3, argv=<optimized out>, 
    argv@entry=0x7ffc50fce198, res_error=res_error@entry=0x7ffc50fce078)
    at /home/smcv/src/flatpak/app/flatpak-main.c:854
#5  0x000055cc22a2ee68 in main (argc=3, argv=0x7ffc50fce198) at /home/smcv/src/flatpak/app/flatpak-main.c:959
```

This seems to be a bug in libappstream rather than a bug in Flatpak. The assertion failure is here:

```
	/* ensure the token cache is created */
[...]

PR #385, opened 2022-02-26
AsComponent: Centralize GOnce guard to fix an assertion failure

smcv:
* trivial: Make as_component_create_token_cache static
    
    It's no longer called from outside AsComponent, so it doesn't need to be
    extern any more.

* AsComponent: Centralize GOnce guard to fix an assertion failure
    
    The token cache only needs to be created once per AsComponent.
    Previously, each of the two callers of as_component_create_token_cache()
    guarded it with an atomic (GOnce) access to priv->token_cache_valid,
    but the implementation had a redundant non-atomic guard. This resulted
    in a GLib assertion failure on leaving the function, because the atomic
    boolean flag indicating whether initialization has completed is not
    meant to be set non-null until g_once_init_leave().
    
    In particular, this resolves test failures in Flatpak 1.13.x
    (which has been ported from libappstream-glib to libappstream) when it
    runs `flatpak search` from its test suite.
    
    Resolves: https://github.com/ximion/appstream/issues/384

### DIFF
--- a/src/as-component-private.h
+++ b/src/as-component-private.h
@@ -67,7 +67,6 @@ const gchar		*as_component_get_architecture (AsComponent *cpt);
 void			 as_component_set_architecture (AsComponent *cpt,
 							const gchar *arch);
 
-void			as_component_create_token_cache (AsComponent *cpt);
 GPtrArray		*as_component_generate_tokens_for (AsComponent *cpt,
 							   AsSearchTokenMatch token_kind);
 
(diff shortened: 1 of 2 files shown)


ximion:
Indeed, and I do think this is exactly the right fix right now - in case the data structure being dynamically allocated causes a problem (I'll need to have a look at what GLib is actually doing there), we could always use a mutex, or maybe better GLib's atomic operations helpers.
Thank you for the patch, sorry for the review delay (vacation had to happen ;-) )

xerexcoded:
AppStream version: 0.15.2 , still does not have this merged , hopefully the next version will resolve the error

PR #392, opened 2022-03-23
Add vcs-browser and contribute URL type

JakobDev:
This PR adds vcs-browser and contribute as URL types as in #374. It was easier and cleaner to make a new PR than to update the old one.

### DIFF
--- a/src/as-enums.h
+++ b/src/as-enums.h
@@ -81,6 +81,8 @@ typedef enum {
  * @AS_URL_KIND_DONATION:	Page with information about how to donate to the project
  * @AS_URL_KIND_TRANSLATE:	Page with instructions on how to translate the project / submit translations.
  * @AS_URL_KIND_CONTACT:	Contact the developers
+ * @AS_URL_KIND_VCS:	Browse the source code
+ * @AS_URL_KIND_CONTRIBUTE:	Help developing
  *
  * The URL type.
  **/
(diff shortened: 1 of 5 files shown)


src/as-enums.h
```diff
@@ -93,6 +95,8 @@ typedef enum {
 	AS_URL_KIND_DONATION,
 	AS_URL_KIND_TRANSLATE,
 	AS_URL_KIND_CONTACT,
+	AS_URL_KIND_VCS,
```
ximion, line 98: This should be `AS_URL_KIND_VCS_BROWSER` shouldn't it?

ximion (changes requested):
Looks good to me, except for the VCS enum name that should go after its string representation, for clarity.

JakobDev:
I had now replaced AS_URL_KIND_VCS with AS_URL_KIND_VCS_BROWSER

ximion (approved):

ximion:

JakobDev:
I informed [GNOME](https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1702) and [KDE](https://bugs.kde.org/show_bug.cgi?id=452054) that the new types has been added. Would be nice, if you can create a new release. It will take some time until the distros ship the new Version, so the Implementation in the Software Centers can be started.

cassidyjames:
Thanks @JakobDev, I'll file the appropriate issues on elementary software.

PR #403, opened 2022-05-20
Validate timestamp

### LINKED ISSUE #402: appstreamcli doesn't validate timestamp
There was a [Bug report](https://github.com/flathub/website/issues/282) for the Flathub website, that the release date, was 52 years ago. I checked the [Appstream file of the App](https://gitlab.com/bit3/whatsappqt/-/blob/master/desktop/share/metainfo/io.bit3.WhatsAppQT.metainfo.xml) which contains this lines:
```xml
<releases>
  <release version="1.0.3" timestamp="2021-11-15"/>
  <release version="1.0.2" timestamp="2021-06-06"/>
  <release version="1.0.1" timestamp="2020-08-09"/>
  <release version="1.0.0" timestamp="2020-08-09"/>
</releases>
```
This file accidentality uses timestamp instead of Date. I downloaded the file to see if appstreamcli find the error. The file pass the validation. I know, that Flathub currently uses appstream-glib, but Flatpak and Flathub working on a switch to libappstream, so it would be nice if you could add a check so these things can be blocked in the future.

JakobDev:
Fixes #402. This fix is a simple if, which doesn't needed much experience in C, so I decried to do the fix myself. I checked for a Timestamp smaller than 3000, if someone has the idea to use the Year as timestamp. I guess no one wants to use 1970/1/1 as release date in the Appstream, so it wouldn't break anything.

### DIFF
--- a/src/as-validator-issue-tag.h
+++ b/src/as-validator-issue-tag.h
@@ -755,7 +755,6 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 
 	{ "release-timestamp-invalid",
 	  AS_ISSUE_SEVERITY_ERROR,
-	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
 	  N_("The release timestamp is invalid."),
 	},
 
--- a/src/as-validator-issue-tag.h
+++ b/src/as-validator-issue-tag.h
@@ -753,6 +753,12 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("The release is missing either the `date` (preferred) or the `timestamp` property."),
 	},
 
+	{ "release-timestamp-invalid",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
+	  N_("The release timestamp is invalid."),
+	},
+
 	{ "artifact-type-invalid",
 	  AS_ISSUE_SEVERITY_ERROR,
 	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
--- a/src/as-validator.c
+++ b/src/as-validator.c
@@ -1689,6 +1689,11 @@ as_validator_check_release (AsValidator *validator, xmlNode *node, AsFormatStyle
 		/* Neither timestamp, nor date property exists */
 		if (timestamp == NULL)
 			as_validator_add_issue (validator, node, "release-time-missing", "date");
+		else {
+			if (atoi(timestamp) <= 3000) {
+				as_validator_add_issue (validator, node, "release-timestamp-invalid", "timestamp");
+			}
+		}
 	}
 	prop = as_xml_get_prop_value (node, "date_eol");
 	if (prop != NULL) {
--- a/src/as-validator.c
+++ b/src/as-validator.c
@@ -1690,9 +1690,9 @@ as_validator_check_release (AsValidator *validator, xmlNode *node, AsFormatStyle
 		if (timestamp == NULL)
 			as_validator_add_issue (validator, node, "release-time-missing", "date");
 		else {
-			/* check if the timestamp is both a number and higher than 3000. The 3000 is used to check, that it is not a year */
 			if (g_ascii_strtoll (timestamp, NULL, 10) < 3000) {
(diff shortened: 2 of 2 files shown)


src/as-validator.c
```diff
@@ -1689,6 +1689,11 @@ as_validator_check_release (AsValidator *validator, xmlNode *node, AsFormatStyle
 		/* Neither timestamp, nor date property exists */
 		if (timestamp == NULL)
 			as_validator_add_issue (validator, node, "release-time-missing", "date");
+		else {
+			if (atoi(timestamp) <= 3000) {
```
ximion, line 1693: This arbitrarily checks for a "new enough" timestamp, it would be better to just check if the string value is a valid integer.
Like this:
```c
gint64 timestamp_int = 0;
errno = 0;
timestamp_int = g_ascii_strtoll (timestamp, NULL, 10);
if (errno != 0 && timestamp_int == 0)
    as_validator_add_issue (validator, node, "release-timestamp-invalid", timestamp);
```

JakobDev: Would not be better to check both: Is it is a number and is it new enough? This would also catch cases where somebody try to use the Year as timestamp.

ximion: On the one hand I don't like it that this would result in false-positives when some project decided to add a long release history going back decades. On the other hand, it is probably not a likely scenario at all.
But if we add a "Software too old" date, we should go all the way, I would likely accept something like this:
```c
/* check if the timestamp is both a number and a time after 1. Jan 2000 GMT+0 */
if (g_ascii_strtoll (timestamp, NULL, 10) < 946684800)
    as_validator_add_issue (validator, node, "release-timestamp-invalid", timestamp);
```

JakobDev: What about using 3000 as above? It would filter out every Year and you can add Dates from the 90s.

ximion: What if someone entered 20220520? ;-)
But okay, you convinced me! But please use `g_ascii_strtoll` (we don't want locale support and this is the much better choice here) and add a comment why 3000 is used to compare against :-)

src/as-validator-issue-tag.h
```diff
@@ -753,6 +753,12 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("The release is missing either the `date` (preferred) or the `timestamp` property."),
 	},
 
+	{ "release-timestamp-invalid",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
```
ximion, line 758: You can drop the translator hint, there are no tag names here.

src/as-validator.c
```diff
@@ -1689,6 +1689,12 @@ as_validator_check_release (AsValidator *validator, xmlNode *node, AsFormatStyle
 		/* Neither timestamp, nor date property exists */
 		if (timestamp == NULL)
 			as_validator_add_issue (validator, node, "release-time-missing", "date");
+		else {
+			/* check if the timestamp is both a number and higher than 3000. The 3000 is used to check, that it is not a year */
+			if (g_ascii_strtoll (timestamp, NULL, 10) < 3000) {
+				as_validator_add_issue (validator, node, "release-timestamp-invalid", "timestamp");
```
ximion, line 1695: The `hint` parameter should be the actual timestamp and not a "timestamp" string, so `as_validator_add_issue (validator, node, "release-timestamp-invalid", timestamp);`

src/as-validator.c
```diff
@@ -1689,6 +1689,12 @@ as_validator_check_release (AsValidator *validator, xmlNode *node, AsFormatStyle
 		/* Neither timestamp, nor date property exists */
```
ximion, line 1689: nitpick: Please move that comment past the `if (timestamp == NULL)` condition:
```c
if (timestamp == NULL) {
    /* Neither timestamp, nor date property exists */
    as_validator_add_issue (validator, node, "release-time-missing", "date");
} else {
    ....
```

ximion (changes requested):
A few minor things, and then this should be good to go!

src/as-validator.c
ximion: You missed this one, I think ;-)

JakobDev: Sorry

ximion:
Thanks for the patch!

ximion (approved):

Issue #335, opened 2021-07-14
display_length size constants don't reflect actual screens


alice-mkh:
```
* xsmall - Very small screens, as used in watches, wearables and other small-display devices (about <= 360px).
* small - Small screens often used in handheld devices, such as phone screens, small phablets (about < 768px).
* medium - Screens in laptops, tablets (about >= 768px)
* large - Bigger computer monitors (about >= 1024px)
* xlarge - Television screens, large projected images (about >= 3840px)
```

There are a few problems here:

* The `small` size is meant to be used on phones, yet all the available phones have `360px` and so fall into `xsmall`
* Meanwhile, `xsmall` doesn't constrain the minimum screen size
* Together this means there's no constant that can work for devices like L5 or PinePhone - `small` means the app cannot run on it, `xsmall` means the app can run on even smaller screens which is not really the case
* `medium` is means for laptops while `large` for bigger computer monitors - the most common laptop resolutions are 1366x768 (mostly low-end), 1920x1080, and there's also 2160x1440@2 which means 1280x720. So laptop sizes can be 720px, 1080px - the former falls into `small`, the latter falls into `large` - so the `medium` description is misleading
* All values correspond to the smallest screen sides, but `xlarge` corresponds to the largest screen side. 4K screens are 3840x2160 - the value there is 2160px so they are still considered large, so `xlarge` only corresponds to 8K screens.

CC @bertob

bertob:
FWIW this is why I always thought the named values were a bad idea. At the end of the day what matters are the actual logical pixel values apps can scale to. This just obscures that, making it easier to get it wrong.

Even if we e.g. change the `small` constant to mean "360 or more" now it may become a problem again in the future when new devices come out that don't neatly fit into these categories...

bertob:
On a practial level, what are our options here?

Can we deprecate/remove the named values altogether? If so, that'd be the best path forward IMO.

If that's not an option, can we add/remove/change the meaning of the various values? If so, to what degree?

ximion:
We can't remove the named values, but we can always change what they mean, as that is exactly their intended purpose (for apps which are fully adaptive and really don't care all that much about explicit sizes, because specifying an absolute size means the app will not show up on a phone that is just a bit too small or a bit too large).
`xsmall` doesn't have a minimum size the same way that `xlarge` doesn't have a maximum size, simply because it basically means "smallest/largest screens we know of".
I'd happily accept a patch for better descriptions and/or rough size recommendations.

[...]

PR #404, opened 2022-05-30
Fix as_display_length_kind_from_px()

JakobDev:
The current implementation returns AS_DISPLAY_LENGTH_KIND_SMALL when px is higher than 360 and AS_DISPLAY_LENGTH_KIND_XSMALL when px is 360 or lower. All other are not returned. This bug was found by LGTM.

### DIFF
--- a/src/as-relation.c
+++ b/src/as-relation.c
@@ -423,17 +423,16 @@ as_display_length_kind_to_px (AsDisplayLengthKind kind)
 AsDisplayLengthKind
 as_display_length_kind_from_px (gint px)
 {
-	if (px <= 360 )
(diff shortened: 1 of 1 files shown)


ximion:
Neat, thanks!
Please keep in mind that we will likely remove the labelled text values in AppStream 1.0, so they should not be used if possible.

JakobDev:
Why did you plan to remove this?

ximion:
Because of https://github.com/ximion/appstream/issues/335
This has been an annoyance from the very start, and others convinced me that trying to define human-readable size classes is a futile exercise.

PR #406, opened 2022-06-03
Add validation for <memory>

JakobDev:
This checks, if the content of <memory> is a number

### DIFF
--- a/src/as-validator-issue-tag.h
+++ b/src/as-validator-issue-tag.h
@@ -342,6 +342,12 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("This `hardware` item contains an invalid value. It should be a Computer Hardware ID (CHID) UUID without braces.")
 	},
 
+	{ "relation-memory-value-invalid",
+	  AS_ISSUE_SEVERITY_WARNING,
+	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
+	  N_("This `memory` item contains an invalid value. It should only contain digits")
+	},
+
 	{ "component-type-invalid",
 	  AS_ISSUE_SEVERITY_ERROR,
 	  N_("The set component type is not a recognized, valid AppStream component type.")
--- a/src/as-validator-issue-tag.h
+++ b/src/as-validator-issue-tag.h
@@ -345,7 +345,7 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	{ "relation-memory-value-invalid",
 	  AS_ISSUE_SEVERITY_WARNING,
 	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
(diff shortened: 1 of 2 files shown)


src/as-validator.c
```diff
@@ -1473,6 +1473,15 @@ as_validator_check_relations (AsValidator *validator,
 				as_validator_add_issue (validator, iter, "relation-hardware-value-invalid", content);
 		}
 
+		/* check memory for sanity */
+		if (item_kind == AS_RELATION_ITEM_KIND_MEMORY) {
+			for (guint i = 0; content[i] != '\0'; i++) {
```
ximion, line 1478: Instead of this loop you could also just do a `if (g_ascii_strtoll (content, NULL, 10) == 0)`, that would be quite a bit more readable.

src/as-validator-issue-tag.h
```diff
@@ -342,6 +342,12 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("This `hardware` item contains an invalid value. It should be a Computer Hardware ID (CHID) UUID without braces.")
 	},
 
+	{ "relation-memory-value-invalid",
+	  AS_ISSUE_SEVERITY_WARNING,
+	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
+	  N_("This `memory` item contains an invalid value. It should only contain digits")
```
ximion, line 348: With the `g_ascii_strtoll` change from below, I would rephrase this as
```
A `memory` item must only contain a non-zero integer value, depicting a system memory size in mebibyte (MiB)
```

ximion (changes requested):

src/as-validator.c
JakobDev: I have tested this. It let thins like `<memory>8GB</memory>` pass and I haven't find a function that checks if a whole string has only digits.

ximion: Right, `strtoll` is actually very forgiving. In this case, please use the (brand new!) `as_str_verify_integer (content, 1, G_MAXINT64))` function here :-)

JakobDev: I have now this code:
```c
/* check memory for sanity */
if (item_kind == AS_RELATION_ITEM_KIND_MEMORY) {
	if (as_str_verify_integer (content, 1, G_MAXINT64)) {
		as_validator_add_issue (validator, iter, "relation-memory-value-invalid", content);
	}
}
```
But it does not work. It didn't catch anything. I had already merged the latest Upstream commits.

ximion: This condition needs to be negated - at the moment you are checking if the value is an integer within the set limits and if that is the case, you emit the issue - you do want to do the exact opposite though ;-)

Also, you can remove the extra braces, braces are only added when multiple lines are within the if condition, or if it is followed by an else branch that also contains multiple statements. In any other case, omitting braces if fine and preferred :-)

ximion (approved):

ximion:

PR #407, opened 2022-06-03
Check if URL type is redefined

JakobDev:
This checks, if a URL type is used more than once. The code is based on c6e5d65db1d3ffda49475061cd1a2a44fec5f243. It is tested and it works. Hope I got everything right.


ximion:
This should definitely be done using a bitfield, or at the very least by using GHashTable as a set, rather than a map. But the bitfield method would be a lot nicer.

JakobDev:
Could you provide a example for bitfield?

ximion:
Check out https://en.wikipedia.org/wiki/Bit_field, but the linked commit that implements this is probably a lot better as a reference. AppStream has a few macros which make bitwise operations easier as well :-)

PR #408, opened 2022-06-03
Update duplicate checking

JakobDev:
This PR adds 2 things:

1. as_validator_check_appear_once() currently allow all tags to have translations. This doesn't make sense for tags like project_license. You can't have one License for English users and one for Spanish users. I added a argument to tell, if translations are allowed or not. If not, the new Issuetag unsupported-translation will be raised.

2. I added the duplicate check to a few more tags e.g. <id>.

Please tell me, if a made a mistake at some point.

### DIFF
--- a/src/as-validator-issue-tag.h
+++ b/src/as-validator-issue-tag.h
@@ -91,6 +91,11 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("The description contains a web URL in plain text. This is not allowed, please use the <url/> tag instead to share links.")
 	},
 
+	{ "unsupported-translation",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("As per AppStream specification, the mention tag doesn't support translations.")
+	},
+
 	{ "tag-duplicated",
 	  AS_ISSUE_SEVERITY_ERROR,
 	  N_("As per AppStream specification, the mentioned tag must only appear once in this context. Having multiple tags of this kind is not valid.")
--- a/src/as-validator.c
+++ b/src/as-validator.c
@@ -882,7 +882,7 @@ as_validator_check_description_tag (AsValidator *validator, xmlNode* node, AsFor
  * as_validator_check_appear_once:
  **/
 static void
-as_validator_check_appear_once (AsValidator *validator, xmlNode *node, GHashTable *known_tags)
+as_validator_check_appear_once (AsValidator *validator, xmlNode *node, GHashTable *known_tags, bool allowTranslations)
 {
 	g_autofree gchar *lang = NULL;
 	gchar *tag_id;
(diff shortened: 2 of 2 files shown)


src/as-validator-issue-tag.h
```diff
@@ -91,6 +91,11 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("The description contains a web URL in plain text. This is not allowed, please use the <url/> tag instead to share links.")
 	},
 
+	{ "unsupported-translation",
```
ximion, line 94: naming: better: `translation-unsupported` or `tag-not-translatable`

src/as-validator-issue-tag.h
```diff
@@ -91,6 +91,11 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
...
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("As per AppStream specification, the mention tag doesn't support translations.")
```
ximion, line 96: typo: better: `This tag is not translatable` (everything is per the specification, so we usually omit overly verbose wording)

src/as-validator.c
```diff
@@ -882,7 +882,7 @@ as_validator_check_description_tag (AsValidator *validator, xmlNode* node, AsFor
  * as_validator_check_appear_once:
  **/
 static void
-as_validator_check_appear_once (AsValidator *validator, xmlNode *node, GHashTable *known_tags)
+as_validator_check_appear_once (AsValidator *validator, xmlNode *node, GHashTable *known_tags, gboolean allowTranslations)
```
ximion, line 885: style: Please use snake_case => `allow_translations` - but for conciseness, I'd probably name this `translatable` or `allow_l10n` (your choice though) :)

src/as-validator.c
```diff
@@ -1960,9 +1965,11 @@ as_validator_validate_component_node (AsValidator *validator, AsContext *ctx, xm
 
 			/* validate the AppStream ID */
 			as_validator_validate_component_id (validator, iter, cpt);
+
```
ximion, line 1968: nit: extra space not needed

src/as-validator.c
```diff
@@ -893,8 +893,13 @@ as_validator_check_appear_once (AsValidator *validator, xmlNode *node, GHashTabl
 	lang = as_xml_get_prop_value (node, "lang");
 	if (lang == NULL)
```
ximion, line 894: style: This *does* need braces now, as the following `else` branch also has them

ximion (changes requested):
This looks pretty good overall, I had just a bunch of style comments and the tag_id that's added to the hash table should be the unlocalized one in case a tag isn't translatable.

ximion (approved):

PR #414, opened 2022-06-23
[qt] Rename Component::requires to avoid clash with C++20

### LINKED ISSUE #342: Build error in C++20 project
qt/component.h has a function named 'requires'. When including this header in a project that is building with C++20 this will result in a compile error since requires is a reserved keyword in C++20

```
In file included from /usr/include/AppStreamQt/pool.h:28,
                 from /home/nico/kde/src/plasma-workspace/applets/kicker/plugin/actionlist.cpp:45:
/usr/include/AppStreamQt/component.h:195:36: error: expected unqualified-id before 'requires'
  195 |         QList<AppStream::Relation> requires() const;
```

nicolasfella:
In C++20 requires is a reserved keyword

This results in a build failure when a C++20 project includes component.h

To avoid the issue the method gets a new name (requiresComponents). The old method is marked as deprecated and hidden when building in C++20 mode

Fixes #342

### DIFF
--- a/qt/component.cpp
+++ b/qt/component.cpp
@@ -361,7 +361,14 @@ void Component::addReplaces(const QString &cid)
     as_component_add_replaces(m_cpt, qPrintable(cid));
 }
 
+#if __cplusplus < 202002L
 QList<Relation> Component::requires() const
+{
+    return requirements();
+}
+#endif
+
+QList<Relation> Component::requirements() const
 {
     QList<AppStream::Relation> res;
 
--- a/qt/component.h
+++ b/qt/component.h
@@ -199,7 +199,10 @@ class APPSTREAMQT_EXPORT Component {
         QStringList replaces() const;
         void addReplaces(const QString& cid);
 
-        QList<AppStream::Relation> requires() const;
+#if __cplusplus < 202002L
+        Q_DECL_DEPRECATED QList<AppStream::Relation> requires() const;
+#endif
+        QList<AppStream::Relation> requirements() const;
         QList<AppStream::Relation> recommends() const;
         QList<AppStream::Relation> supports() const;
         void addRelation(const AppStream::Relation &relation);


qt/component.cpp
```diff
@@ -364,7 +364,14 @@ QList<Relation> Component::recommends() const
     return res;
 }
 
+#if __cplusplus < 202002L
```
Vogtinator, line 367: Maybe add a `#warning` in an `#endif` so that if AppStreamQt is built with C++20, it will be ABI incompatible?

ximion:
Thank you for the patch! I do think `requiresComponents` is a terrible name though, because pretty much anything can be required, including screen dimensions, memory, mime-handlers or binaries, and not just components. So that name would be very misleading.
This patch would also make the library break ABI depending on which C++ standard it was compiled with, which I am not really happy with...

hughsie:
> In C++20 requires is a reserved keyword

That kind of change seems somewhat, well, hostile to your existing userbase...

jwakely:
You can use asm to create a symbol alias to continue exporting a symbol with the old name, so there's no ABI break.

marxin:
https://gcc.gnu.org/onlinedocs/gcc-12.1.0/gcc/Common-Function-Attributes.html#Common-Function-Attributes
seek for `alias ("target")`

jwakely:
> That kind of change seems somewhat, well, hostile to your existing userbase...

The `requires` keyword was first added to the C++ working draft in 2008, and although it was removed again before C++11 was published, the fact that Concepts were coming back *at some point* was always certain. It was a poor choice of identifier any time after 2008.

ximion:
> Wouldn't requirements be better anyway?

That's probably what I would go with... It's called "requires" because that's what the tag name is in metadata and it gives well with the existing replaces/extends/recommends/supports - to be consistent those would also need to be named "recommendations", "replacement_for", "supporting" etc, which is a bit weird.
I think requirements is fine here even though it breaks consistency, just to get rid of the name clash with "requires" becoming a reserved word.

nicolasfella:
Renamed it to requirements

nicolasfella:
> https://gcc.gnu.org/onlinedocs/gcc-12.1.0/gcc/Common-Function-Attributes.html#Common-Function-Attributes seek for `alias ("target")`

I'm not sure I follow. That site says "The alias attribute causes the declaration to be emitted as an alias for another symbol, which must have been previously declared with the same type". So in order to create an alias for it I have to declare it, but I can't, because the name is a keyword. Or am I misunderstanding something?

jwakely:
You can define it in a translation unit compiled with C++17 (or earlier) where it's not a keyword, and make it an alias to the new name. Or just make it call the new function.

You can also declare it as an `extern "C"` symbol using its mangled name and alias that the the new mangled name, but that's less portable, and sneaky.

Issue #343, opened 2021-08-10
Add <internet/> element to <requires>/<recommends>


pwithnall:
It would be interesting to expose metadata about whether an application requires or recommends internet access to work, or whether it can work fully offline. For users who have limited or sporadic internet access this would allow them to choose applications which work well in those situations. Such users could, for example, be on a mobile data plan for their internet (which runs out), be travelling with patchy signal, or be on a satellite internet connection which has high enough latency to make it annoying to use.

This is something we’d be interested in presenting in the hardware context tile in gnome-software, which currently summarises the existing `<requires>`/`<recommends>` elements.

Suggested XML would be an `<internet>` element whose value is the minimum internet speed the app requires, in megabits/second. A value of `0` would mean that the internet is required, but a specific speed isn’t needed. An `offline="yes|no"` element would indicate whether the app can work when offline (by pre-caching content).

Using `<internet>` inside `<requires>` would indicate that the app is useless without some internet access at some point. Using `<internet>` inside `<recommends>` would indicate that the app can largely work without the internet, but its experience would be enhanced by internet access.

Examples:
```
<requires><internet>0</internet></requires> <!-- App can’t function without some form of internet connection -->
<recommends><internet>2</internet></requires> <!-- App works better if it has a >=2MBit/s internet connection -->
<requires><internet offline="yes">0</internet></requires> <!-- App can work if offline, but needs an internet connection at some point, perhaps when first run to download and cache some data -->
```

How does that sound?

/cc @wjt because you had some ideas about this

wjt:
The speed case is interesting, I hadn't thought about that. The cases I was interested in are:

1. App does not need an internet connection. (Example: Encyclopedia on Endless OS, [Contrast](https://flathub.org/apps/details/org.gnome.design.Contrast), [Teleport](https://flathub.org/apps/details/com.frac_tion.teleport))
2. App is enhanced by an internet connection but doesn't need one. (Example: web browsers, which can open local content; games like OpenTTD which are *usable* offline but really you want to go online at least once to download some campaigns; Vorta, which can be used to back up to local disks or remote disks.)
3. App is useless without an internet connection. (Example: Dropbox, Spotify, etc.)

(I mention Teleport as an example because it has the `--share=network` Flatpak permission. Absence of this permission implies the app is usable without an internet connection, but its presence does not mean the app needs an internet connection. `man flatpak-build-finish` also documents that `--share=network` is needed to fully use Bluetooth.)

[...]

PR #421, opened 2022-07-23
Add an <internet/> kind to requires/recommends/supports

pwithnall:
See the commit messages. This adds a new `<internet/>` kind to requires/recommends/supports, both in the spec and in libappstream. It includes unit tests.

Fixes: #343

### DIFF
--- a/docs/xml/metainfo-component.xml
+++ b/docs/xml/metainfo-component.xml
@@ -1106,6 +1106,58 @@
 				</listitem>
 			</varlistentry>
 
+			<varlistentry id="tag-relations-internet">
+				<term>&lt;internet/&gt;</term>
+				<listitem>
+				<para>
+					Require, recommend or support connectivity to the internet. The value of this item one of the following:
+				</para>
+				<itemizedlist>
+					<listitem><para><code>always</code> - Needs internet connectivity to work. If used in a <literal>recommends</literal> element, then this indicates that the app can work without internet, but the experience will be significantly degraded.</para></listitem>
+					<listitem><para><code>offline-only</code> - Never uses the internet, even if it’s available.</para></listitem>
+					<listitem><para><code>first-run</code> - Uses the internet the first time the application is run, but not normally afterwards.</para></listitem>
+				</itemizedlist>
+				<para>
+					If the value of <literal>&lt;internet/&gt;</literal> is not <literal>offline-only</literal>, the <literal>bandwidth_mbitps</literal> attribute can be
+					set to a bandwidth (in Mbit/s) which is the minimum internet bandwidth needed for the application to be usable. If this attribute is not set, it’s
+					assumed that the application is usable with all internet connections.
+				</para>
+				<para>
+					Examples:
+				</para>
+				<programlisting language="XML"><![CDATA[<!-- always needs the internet -->
+<requires>
+  <internet>always</internet>
+</requires>
+
+<!-- always needs the internet and has a degraded experience if it’s not at least 2Mbit/s -->
+<requires>
+  <internet bandwidth_mbitps="2">always</internet>
+</requires>
+
+<!-- never uses the internet, even if available -->
+<requires>
+  <internet>never</internet>
+</requires>
+
+<!-- requires the internet on first run -->
+<requires>
+  <internet>first-run</internet>
+</requires>
+
+<!-- can work without the internet, but with a degraded experience -->
+<recommends>
+  <internet>always</internet>
+</recommends>
+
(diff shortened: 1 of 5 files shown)


pwithnall:
I’m going to be away for a few months from basically now, so won’t be able to respond to review feedback on this after tomorrow or so. Please feel free to update/change/drop/close/adapt the MR in whatever way seems most appropriate.

One thing you might want to think about in particular is the `bandwidth_mbitps` attribute, which was more of a question mark in the discussion on #343 than the rest of the new bits in the spec. It should be fairly straightforward to drop `bandwidth_mbitps` from this MR if you don’t think it’s ready to add that now. It can always be added to the spec in future if people end up thinking that’s a good idea.

docs/xml/metainfo-component.xml
```diff
@@ -1106,6 +1106,58 @@
...
 
+			<varlistentry id="tag-relations-internet">
+				<term>&lt;internet/&gt;</term>
+				<listitem>
+				<para>
+					Require, recommend or support connectivity to the internet. The value of this item one of the following:
+				</para>
+				<itemizedlist>
+					<listitem><para><code>always</code> - Needs internet connectivity to work. If used in a <literal>recommends</literal> element, then this indicates that the app can work without internet, but the experience will be significantly degraded.</para></listitem>
+					<listitem><para><code>offline-only</code> - Never uses the internet, even if it’s available.</para></listitem>
```
ximion, line 1117: What is the value in case an application *can* use the internet to enhance the experience, but otherwise does not need to and would also work perfectly fine offline? (e.g. a media player fetching cover art for songs from the web, but which would play them perfectly find just as well w/o internet connection)
Would that be maybe:
```xml
<supports>
  <internet>always</internet>
</supports>
```

src/as-relation.c
```diff
@@ -1044,6 +1184,16 @@ as_relation_load_from_xml (AsRelation *relation, AsContext *ctx, xmlNode *node,
 		g_autofree gchar *side_str = (gchar*) xmlGetProp (node, (xmlChar*) "side");
 		priv->display_side_kind = as_display_side_kind_from_string (side_str);
 
+		g_free (priv->version);
```
ximion, line 1187: I know it's not your code, but what about `g_free (g_steal_pointer (&priv->version))` or `as_relation_set_version (relation, NULL)`?

src/as-relation.c
```diff
@@ -1044,6 +1184,16 @@ as_relation_load_from_xml (AsRelation *relation, AsContext *ctx, xmlNode *node,
...
+		priv->version = NULL;
+	} else if (priv->item_kind == AS_RELATION_ITEM_KIND_INTERNET) {
+		g_autofree gchar *bandwidth_str = (gchar*) xmlGetProp (node, (xmlChar*) "bandwidth_mbitps");
```
ximion, line 1190: `g_autofree gchar *bandwidth_str = as_xml_get_prop_value (node, "bandwidth_mbitps");`
(a bit nicer to read)

src/as-relation.c
```diff
@@ -1101,6 +1251,11 @@ as_relation_to_xml_node (AsRelation *relation, AsContext *ctx, xmlNode *root)
 			     (xmlChar*) as_relation_item_kind_to_string (priv->item_kind),
 			     (xmlChar*) as_control_kind_to_string (as_relation_get_value_control_kind (relation)));
 
+	} else if (priv->item_kind == AS_RELATION_ITEM_KIND_INTERNET) {
+		n = xmlNewTextChild (root, NULL,
```
ximion, line 1255: You could use `as_xml_add_text_node` here, but that's optional :-)

ximion (changes requested):

ximion:
Looks really good, all the comments I had are very minor. The only thing I think we need to answer is the "app that can use internet but doesn't need to" case.
I initially thought of dropping the `bandwidth_mbitps` stuff, but tbh I think it looks fine and I would be okay with adding it now *if* you/your apps/your client would use it. If it has zero users, then I would be a lot more hesitant with including it.

All in all, thank you a lot for the really good patch and sorry for the delayed reply as well as causing a merge conflict by converting the AsRelation code to using the XML helpers ;-) (they don't do much except for making the C code easier to read).

JakobDev:
You should probably explicit mention that using it multiple times is supported. Something like this:
```xml
<requires>
  <internet>first-run</internet>
</requires>
<recommends>
  <internet>always</internet>
</recommends>
```
This is to avoid confusion in the implementations.

ximion:
Closing in favour of https://github.com/ximion/appstream/pull/425

PR #422, opened 2022-07-24
data: use lxml to get the XDG categories

pinotree:
Use lxml to get the list of XDG categories from the DocBook file using
XPath queries; this simplifies the code a bit, and makes it more robust.

There is no change in the extracted data.

### DIFF
--- a/data/update-data.py
+++ b/data/update-data.py
@@ -9,17 +9,21 @@
 #
 
 import os
+import io
 import json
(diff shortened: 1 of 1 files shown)


ximion:
I think the original intent was to only use Python stdlib stuff... But XPath queries are so much nicer, thank you for improving this piece of code! :-)

PR #425, opened 2022-08-20
Add an <internet/> kind to requires/recommends/supports (2)

ximion:
This is a fixed up version of https://github.com/ximion/appstream/pull/421 that is ready to merge (due to @pwithnall being on vacation and there's no reason to leave this 99.9% completed feature out of the next release).

@cassidyjames @JakobDev Please have a look if you see any issues, otherwise I hope I can merge this in time for the release that will likely happen in the evening next Monday.

PR #426, opened 2022-08-22
Add an <internet/> kind to requires/recommends/supports /3

ximion:
This is a version of https://github.com/ximion/appstream/pull/425 with some minor fixes and added validation.

### DIFF
--- a/docs/xml/metainfo-component.xml
+++ b/docs/xml/metainfo-component.xml
@@ -1106,6 +1106,58 @@
 				</listitem>
 			</varlistentry>
 
+			<varlistentry id="tag-relations-internet">
+				<term>&lt;internet/&gt;</term>
+				<listitem>
+				<para>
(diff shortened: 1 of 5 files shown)


JakobDev:
Maybe you should also add a example for offline-only, so they User know in which relations-tag he should put it.

ximion:
This is shaping up to be the most extensive example block in the whole specification :-D - I normally like to keep the examples quite brief, as information should be derivable from the text, but I also do see the point here of adding an example for it in a `supports` block for even more clarity.

JakobDev:
I found this in the documentation as exmaple:
```xml
<!-- never uses the internet, even if available -->
<requires>
  <internet>never</internet>
</requires>
```
This is invalid and should be offline-only.

ximion:
Ha! Well spotted! Thanks for having another look, it's good to not make a release with these obvious but very confusing errors (and I didn't spot that when reviewing the original PR!)

PR #430, opened 2022-08-25
qt: Add API for ContentRating descriptions

aleixpol:


ximion:
I think this patch is okay, but wouldn't you also need a wrapper for `as_content_rating_get_rating_ids` to make efficient use of this API?

ximion:
Looks good! I do wonder now whether there will be a time where you'll also want a `attributeDescription(id, value)` API that isn't dependent on having a ContentRating object, like the C API has. But we could add something like that later, in that case... (or of course follow the C API and have these functions separate, but tbh I don't mind that `description()` convenience function :)

ximion:
> I am not sure in which case we'd need a description without an ContentRatings object.

Mostly useful if you want to contruct some kind of overview page :-)

PR #432, opened 2022-09-19
ascli_refresh_cache: fix NULL pointer dereference

r-ricci:
The command `appstreamcli refresh-cache --cachepath /tmp/cache` segfaults. In this code path, `ret` (initialized to `FALSE`) is not updated when `as_pool_load` returns successfully and sets the `error` pointer to NULL. Then the code tries to print `error->message`.


ximion:
Nice, thank you for the patch! I am honestly surprised that static analysis didn't catch that before!

PR #441, opened 2022-10-25
Addresses some problems found while trying to use Pool::componentsByCategories

aleixpol:
It was crashing because the list was not getting populated properly. This changes does so and addresses a couple of extra problems found while doing this as seen in their respective commits2.

### DIFF
--- a/qt/chelpers.h
+++ b/qt/chelpers.h
@@ -77,12 +77,13 @@ inline QStringList valueWrap(GList *list)
 
 inline char ** stringListToCharArray(const QStringList& list)
 {
-    char **array = (char**) g_malloc(sizeof(char*) * list.size());
+    char **array = (char**) g_malloc(sizeof(char*) * list.size() + 1);
     for (int i = 0; i < list.size(); ++i) {
         const QByteArray string = list[i].toLocal8Bit();
         array[i] = (char*) g_malloc(sizeof(char) * (string.size() + 1));
         strcpy(array[i], string.constData());
     }
+    array[list.size()] = nullptr;
     return array;
 }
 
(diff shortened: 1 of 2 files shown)


ximion:
The patches make sense to me except for the `qt: Fix Pool::componentsByCategories` one - which bug does that one fix?
Using `g_new0` already ensures that the resulting memory is nulled, so no need to do that explicitly.
Did the strings from `categories` somehow have a shorter life than the surrounding container?

aleixpol:
> Did the strings from `categories` somehow have a shorter life than the surrounding container?

Note qPrintable is the result of .toUtf8().constData(). We could consider splitting it into two loops, one converting to QByteArray and the other creating the `gchar **`. We'd save a few memcpy() calls.

aleixpol:
If I don't have this commit I sometimes get this error message:
`** (process:152147): WARNING **: 15:14:03.557: '' is not a valid XDG category name, search results might be invalid or empty.`

I've changed the disputed commit to the slightly more verbose but correct and efficient version.

aleixpol:
Ping?

ximion:
> If I don't have this commit I sometimes get this error message: `** (process:152147): WARNING **: 15:14:03.557: '' is not a valid XDG category name, search results might be invalid or empty.`

Really strange! This does point at some kind of lifetime issue though.

> I've changed the disputed commit to the slightly more verbose but correct and efficient version.

I would have even been fine with the previous version, but I do need to understand the rationale behind a change (merging blindly without understanding the cause will eventually haunt you).
But the new version is fine too :-)

Thank you for the patch!

PR #444, opened 2022-10-29
qt: Port away from deprecated foreach

nicolasfella:


ximion:
I didn't know Qt's foreach was deprecated - that's actually really great! :-)
Thanks for the patch!

PR #448, opened 2022-11-17
Improve GObject property implementation in AsReview

pwithnall:
See the commit messages for details.

The bulk of the changes here are to correctly emit `GObject::notify` when properties on `AsReview` are changed. The rest of the changes are mostly best practice / cleanup.

### DIFF
--- a/src/as-review.c
+++ b/src/as-review.c
@@ -52,9 +52,8 @@ typedef struct
 	GHashTable		*metadata; /* GRefString : GRefString */
 } AsReviewPrivate;
 
-enum {
-	PROP_0,
-	PROP_ID,
+typedef enum {
+	PROP_ID = 1,
 	PROP_SUMMARY,
 	PROP_DESCRIPTION,
 	PROP_LOCALE,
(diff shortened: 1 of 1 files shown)


pwithnall:
Some improvements to this code which could be made in future, but which I didn’t make in this PR because I ran out of time:
 * Add an `AsReview::metadata-changed` signal for when `as_review_add_metadata()` is called. Or expose the metadata map as an `AsReview:metadata` property.
 * Change `as_review_load_from_{xml,yaml}()` to construct the object from scratch (rather than having them called on a newly-created `AsReview`) so that they can construct it immutably and avoid emitting N `GObject::notify` signals for N properties set at construction time. That’s an API break though.

ximion:
This looks good! Thank you for the patch!

ximion:
I do intend to release AppStream 1.0 probably next year, which will contain a bunch of API breaks. I am sort of dreading this a bit, since quite a bit of code will need adjustments then, but it's a necessary step to keep this project easy to maintain. So, when this happens we can have any API break that would be useful to you. If you want to, you can add API adjustment plans for 1.0 to [docs/API-TODO.md](https://github.com/ximion/appstream/blob/master/docs/API-TODO.md) as well, so we remember them when we can make those changes :-)

I also wonder whether it would make sense to add the fetch & parse ODRS code into AppStream as well, so it's not duplicated for every software center. Possibly as a new library, as this would likely pull in JSON parsing as well...

pwithnall:
> I also wonder whether it would make sense to add the fetch & parse ODRS code into AppStream as well, so it's not duplicated for every software center. Possibly as a new library, as this would likely pull in JSON parsing as well...

Probably best as a new library, to keep the dependencies separate and make API stability more manageable.

PR #450, opened 2022-11-24
Improve Screenshot validation

JakobDev:
I added the following checks:

- At least one Screenshot must be a default
- Width and Height must be Integers
- `source` and `thumbnail` are the only allowed values for a Image type
- If a Image type is  `thumbnail`, it must have 'width' and 'height'
- Only one Image with type `source` per language is allowed for each Screenshot
- If a Screenshot is a Image, it must contain a source Image

### DIFF
--- a/src/as-validator-issue-tag.h
+++ b/src/as-validator-issue-tag.h
@@ -206,7 +206,7 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 
 	{ "screenshot-invalid-width",
 	  AS_ISSUE_SEVERITY_ERROR,
-	  N_("The width property must be a integer")
+	  N_("The width property must be a positive integer")
 	},
 
 	{ "screenshot-invalid-height",
@@ -216,27 +216,27 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 
 	{ "screenshot-image-invalid-type",
 	  AS_ISSUE_SEVERITY_ERROR,
-	  N_("The type of the image is neither source or thumbnail")
+	  N_("The type of the image is neither `source` or `thumbnail`")
 	},
 
 	{ "screenshot-image-missing-width",
-	  AS_ISSUE_SEVERITY_ERROR,
-	  N_("The width property must be present if the type is thumbnail")
+	  AS_ISSUE_SEVERITY_WARNING,
+	  N_("The `width` property must be present if the type is `thumbnail`")
 	},
 
 	{ "screenshot-image-missing-height",
-	  AS_ISSUE_SEVERITY_ERROR,
-	  N_("The height property must be present if the type is thumbnail")
+	  AS_ISSUE_SEVERITY_WARNING,
+	  N_("The `height` property must be present if the type is thumbnail")
 	},
 
 	{ "screenshot-image-source-duplicated",
 	  AS_ISSUE_SEVERITY_ERROR,
-	  N_("The source type can only be used once for a image and a language")
+	  N_("The 'source' type can only be used once for a image and a language")
 	},
 
 	{ "screenshot-image-source-missing",
 	  AS_ISSUE_SEVERITY_ERROR,
-	  N_("A screenshot must have at leat one image with the source type")
+	  N_("A screenshot must have at least one image of type `source`")
 	},
 
 	{ "screenshot-image-not-found",
@@ -310,7 +310,7 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 
 	{ "screenshot-default-missing",
 	  AS_ISSUE_SEVERITY_ERROR,
-	  N_("There is no defualt screenshot")
+	  N_("A default screenshot is missing.")
 	},
 
 	{ "relation-invalid-tag",
--- a/src/as-validator-issue-tag.h
+++ b/src/as-validator-issue-tag.h
@@ -204,6 +204,41 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("The update-contact does not appear to be a valid email address (escaping of `@` is only allowed as `_at_` or `_AT_`).")
 	},
 
+	{ "screenshot-invalid-width",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("The width property must be a integer")
+	},
+
+	{ "screenshot-invalid-height",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("The width property must be a integer")
+	},
+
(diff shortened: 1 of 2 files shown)


src/as-validator-issue-tag.h
```diff
@@ -204,6 +204,41 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("The update-contact does not appear to be a valid email address (escaping of `@` is only allowed as `_at_` or `_AT_`).")
 	},
 
+	{ "screenshot-invalid-width",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("The width property must be a integer")
```
ximion, line 209: nit: If you want to be super pedantic: "must be a *positive* integer" ;-)

src/as-validator-issue-tag.h
```diff
@@ -204,6 +204,41 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
...
+	},
+
+	{ "screenshot-invalid-height",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("The width property must be a integer")
+	},
+
+	{ "screenshot-image-invalid-type",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("The type of the image is neither source or thumbnail")
```
ximion, line 219: Please put the values like `source` in backticks, and add a hint for translators to not translate these values (see some of the other entries, you can just copy from there).

src/as-validator-issue-tag.h
```diff
@@ -204,6 +204,41 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
...
+	},
+
+	{ "screenshot-image-missing-width",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("The width property must be present if the type is thumbnail")
```
ximion, line 224: ...same for things like `width` if used as meaning the exact, literal name of a tag or property (put in backticks, add translator hint).

src/as-validator-issue-tag.h
```diff
@@ -204,6 +204,41 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
...
+	},
+
+	{ "screenshot-image-source-duplicated",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("The source type can only be used once for a image and a language")
+	},
+
+	{ "screenshot-image-source-missing",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("A screenshot must have at leat one image with the source type")
```
ximion, line 239: typo: least
Maybe: "at least one image of type \`source\`"

src/as-validator-issue-tag.h
```diff
@@ -273,6 +308,11 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("The default screenshot of a software component must not be a video. Use a static image as default screenshot and set the video as a secondary screenshot.")
 	},
 
+	{ "screenshot-default-missing",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("There is no defualt screenshot")
```
ximion, line 313: typo: default
Maybe: "A default screenshot is missing."

src/as-validator.c
```diff
@@ -1228,20 +1228,27 @@ as_validator_check_tags (AsValidator *validator, xmlNode *node)
...
+		g_autoptr(GHashTable) known_source_lang = NULL;
+
+		known_source_lang = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
 
 		if (iter->type != XML_ELEMENT_NODE)
 			continue;
 
 		scr_kind_str = as_xml_get_prop_value (iter, "type");
-		if (g_strcmp0 (scr_kind_str, "default") == 0)
+		if (g_strcmp0 (scr_kind_str, "default") == 0) {
```
ximion, line 1248: Please don't do this. Use `as_screenshot_kind_from_string` and check whether `as_screenshot_kind_from_string (scr_kind_str) == AS_SCREENSHOT_KIND_DEFAULT` (more consistent, easier for refactoring, if you get an `UNKNOWN` you can already raise a validation warning.

src/as-validator-issue-tag.h
```diff
@@ -204,6 +204,41 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
...
+	  N_("The width property must be a integer")
+	},
+
+	{ "screenshot-image-invalid-type",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("The type of the image is neither source or thumbnail")
+	},
+
+	{ "screenshot-image-missing-width",
+	  AS_ISSUE_SEVERITY_ERROR,
```
ximion, line 223: Should probably be a WARNING, the data isn't completely broken if this is missing (warnings will still make validation fail).

src/as-validator.c
```diff
@@ -1257,10 +1264,69 @@ as_validator_check_screenshots (AsValidator *validator, xmlNode *node, AsCompone
 			if (iter2->type != XML_ELEMENT_NODE)
 				continue;
 
+			g_autofree gchar *screenshot_width = NULL;
+			g_autofree gchar *screenshot_height = NULL;
```
ximion, line 1268: Please don't add declarations after statements for now...

src/as-validator.c
```diff
@@ -1257,10 +1264,69 @@ as_validator_check_screenshots (AsValidator *validator, xmlNode *node, AsCompone
...
+
 			if (g_strcmp0 (node_name, "image") == 0) {
 				g_autofree gchar *image_url = as_strstripnl ((gchar*) xmlNodeGetContent (iter2));
+				g_autofree gchar *image_type = NULL;
 
 				image_found = TRUE;
+
+				image_type = as_xml_get_prop_value (iter2, "type");
+				if (image_type == NULL)
+					image_type = "source";
```
ximion, line 1290: Please use `as_screenshot_kind_from_string`, that will do the right thing here and you get an enum value to compare in the subsequent lines, instead of comparing strings.

ximion (changes requested):
Thanks for the patch! In general those are useful improvements, I just had a few comments on the code itself that need to be addressed before merging this (a rebase on master might also be useful, a lot of previous warnings are now strict errors).

ximion:
There will be a release today/tomorrow, would be nice to get this into the last release of this year :-)

ximion:
Can you also rebase this patch on master? Currently it can not be merged easily due to conflicts.

ximion:
I merged this and fixed all mentioned issues in a follow-up commit. Thank you for the patch!

PR #452, opened 2022-12-13
Add validation for custom tag

JakobDev:


src/as-validator-issue-tag.h
```diff
@@ -921,6 +921,26 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("This color is not a valid HTML color code."),
 	},
 
+	{ "custom-invalid-tag",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("The custom tags can only contains value tags"),
```
ximion, line 926: Again, any literal should be in backticks so translators don't translate it by accident.

src/as-validator-issue-tag.h
```diff
@@ -921,6 +921,26 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
...
+	},
+
+	{ "custom-key-missing",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("This custom tag is missing the 'key attribute"),
```
ximion, line 931: => the \`key\` attribute

src/as-validator-issue-tag.h
```diff
@@ -921,6 +921,26 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
...
+	},
+
+	{ "custom-key-duplicated",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("A key can only be used once"),
+	},
+
+	{ "custom-value-missing",
+	  AS_ISSUE_SEVERITY_ERROR,
```
ximion, line 940: Unless this causes problems, making this a WARNING or possibly even INFO might be better. That allows some flexibility for writers of custom data (e.g. to have placeholders that are ignored deliberately).

src/as-validator.c
```diff
@@ -1924,6 +1924,49 @@ as_validator_check_branding (AsValidator *validator, xmlNode *node)
...
 }
 
+/**
+ * as_validator_check_custom:
+ **/
+static void
+as_validator_check_custom (AsValidator *validator, xmlNode *node)
+{
+	g_autoptr(GHashTable) known_keys = NULL;
+	known_keys = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
```
ximion, line 1934: nit: This could be one line with the one above it.

src/as-validator.c
```diff
@@ -1924,6 +1924,49 @@ as_validator_check_branding (AsValidator *validator, xmlNode *node)
...
+
+	for (xmlNode *iter = node->children; iter != NULL; iter = iter->next) {
+		if (iter->type != XML_ELEMENT_NODE)
+			continue;
+
+		g_autofree gchar *node_content = NULL;
+		const gchar *node_name;
```
ximion, line 1941: Please don't mix declarations and statements.

src/as-validator.c
```diff
@@ -1924,6 +1924,49 @@ as_validator_check_branding (AsValidator *validator, xmlNode *node)
...
+
+		node_name = (const gchar*) iter->name;
+
+		if (g_strcmp0 (node_name, "value") != 0) {
+			as_validator_add_issue (validator, iter, "custom-invalid-tag", node_name);
+			continue;
+		}
+
+		gchar *key_name = NULL;
+		key_name = as_xml_get_prop_value (iter, "key");
```
ximion, line 1951: This is missing a `g_autofree`

src/as-validator.c
```diff
@@ -1924,6 +1924,49 @@ as_validator_check_branding (AsValidator *validator, xmlNode *node)
...
+			as_validator_add_issue (validator, iter, "custom-key-missing", NULL);
+			continue;
+		}
+
+		if (g_hash_table_contains (known_keys, key_name))
+			as_validator_add_issue (validator, iter, "custom-key-duplicated", key_name);
+		else
+			g_hash_table_add (known_keys, key_name);
+
+		node_content = (gchar*) xmlNodeGetContent (iter);
```
ximion, line 1963: Please use either `as_xml_get_node_value` or `as_xml_get_node_value_raw`.
You probably want the former and then check for NULL.

ximion (changes requested):
Some comments.
It would actually be super nice if you could also add some basic tests for both patches (see for example `test_validator_overrides` in `tests/test-validate.c`).
For this particular part of AppStream I don't consider this to be critical, but having tests for new features would certainly be nice, to ensure they continue to work :-)

ximion:
There will be a release today/tomorrow, would be nice to get this into the last release of this year :-)

src/as-validator.c
JakobDev: Using this here, causes a crash when added to the table. Probably because it is out of scope.

src/as-validator.c
JakobDev: This unfortunately didn't work
```c
node_content =  as_xml_get_node_value (iter);
if (node_content == NULL) {
    as_validator_add_issue (validator, iter, "custom-value-missing", NULL);
}
```

src/as-validator.c
ximion: You need to change the `g_hash_table_add` line to `g_hash_table_add (known_keys, g_steal_pointer (&key_name));` as well for this to work without a crash.

src/as-validator.c
ximion: What do you mean it did not work?
This code should be fine:
```c
node_content =  as_xml_get_node_value (iter);
if (node_content == NULL || node_content[0] == '\0')
    as_validator_add_issue (validator, iter, "custom-value-missing", NULL);
```

ximion (changes requested):

ximion:
Since the release is imminent, I just addressed the issues for you. Kind of an early christmas present ^^
Thank you for the patch, it's useful to have this additional validation!

sp1ritCS:
This patch has seemingly caused applications using the `Purism::form_factor` custom key in their appstream manifests to fail the appstreamcli validate check, when an application supports more than one form factor. Is this intentional?

https://aplazas.pages.gitlab.gnome.org/blog/blog/2020/10/15/specify-form-factors-in-your-librem-5-apps.html

ximion:
Yes, yikes... The way the keys re used there simply will not work, you can't have two identical keys and assume the values are collected in a list.
So, the validator is correct here.

Maybe simply go the more modern route and set a supported minium display size: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-relations-display_length
As well as a supports tag for touch control: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-relations-control

PR #457, opened 2023-01-20
pool: Offer API to look up by bundles

aleixpol:
It's often that we will need to look up packages by their name. Include some API that helps us do it.

WIP because it would be necessary to look up by e.g. `org.kde.dolphin` too, not just `org.kde.dolphin/x86_64/master`. Some idea on how to do this would be useful.

Context for this change is here:
https://invent.kde.org/plasma/discover/-/merge_requests/455

### DIFF
--- a/qt/pool.cpp
+++ b/qt/pool.cpp
@@ -175,6 +175,14 @@ QList<Component> Pool::componentsByExtends(const QString &extendedId) const
                                                              qPrintable(extendedId)));
 }
 
+QList<Component> Pool::componentsByBundleId(Bundle::Kind kind, const QString &extendedId, bool matchPrefix) const
+{
+    return cptArrayToQList(as_pool_get_components_by_bundle_id(d->pool,
+                                                               static_cast<AsBundleKind>(kind),
+                                                               qPrintable(extendedId),
+                                                               matchPrefix));
+}
+
 QList<AppStream::Component> Pool::search(const QString& term) const
 {
     return cptArrayToQList(as_pool_search(d->pool, qPrintable(term)));
--- a/src/as-cache.c
+++ b/src/as-cache.c
@@ -1812,6 +1812,36 @@ as_cache_get_components_by_launchable (AsCache *cache, AsLaunchableKind kind, co
 					  error);
 }
 
+/**
+ * as_cache_get_components_by_bundle_id:
+ * @cache: An instance of #AsCache.
+ * @kind: Type of the bundle.
+ * @id: ID of the bundle.
+ * @error: A #GError or %NULL.
+ *
+ * Get components which are provided by a bundle with the given kind and ID.
+ *
+ * Returns: (transfer full): An array of #AsComponent
+ */
+
+GPtrArray*
+as_cache_get_components_by_bundle_id (AsCache *cache, AsBundleKind kind, const gchar *id, gboolean match_prefix, GError **error)
+{
+	g_auto(XbQueryContext) context = XB_QUERY_CONTEXT_INIT ();
+	g_autofree gchar *xpath = NULL;
+	const char *match = match_prefix ? "components/component/bundle[@type='%s'][starts-with(text(), ?)]/.."
+					 : "components/component/bundle[@type='%s'][text()=?]/..";
+	xb_value_bindings_bind_str (xb_query_context_get_bindings (&context),
+				    0, id, NULL);
+	xpath = g_strdup_printf (match, as_bundle_kind_to_string (kind));
+	return as_cache_query_components (cache,
+					  xpath,
+					  &context,
+					  0,
+					  FALSE,
+					  error);
+}
+
 typedef struct {
 	AsSearchTokenMatch	match_value;
 	XbQuery			*query;
--- a/src/as-cache.h
+++ b/src/as-cache.h
@@ -182,6 +182,12 @@ GPtrArray		*as_cache_get_components_by_launchable (AsCache *cache,
 								const gchar *id,
 								GError **error);
 
+GPtrArray		*as_cache_get_components_by_bundle_id (AsCache *cache,
+								AsBundleKind kind,
+								const gchar *id,
+								gboolean match_prefix,
+								GError **error);
+
 GPtrArray		*as_cache_search (AsCache *cache,
 					  const gchar * const *terms,
 					  gboolean sort,
--- a/src/as-pool.c
+++ b/src/as-pool.c
@@ -2213,6 +2213,37 @@ as_pool_get_components_by_extends (AsPool *pool, const gchar *extended_id)
 }
 
 /**
+ * as_pool_get_components_by_bundle_id: (skip)
+ * @pool: An instance of #AsPool.
+ * @kind: The kind of the bundle we are looking for
+ * @bundle_id: The bundle ID to match, as specified in #AsBundle
+ * @match_prefix: If the provided bundle id is the prefix or it needs to match exactly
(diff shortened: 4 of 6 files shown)


aleixpol:
No, I do not want the component id, just the bundle id. In fact, part of the problem is that sometimes we know the flatpak bundle id but not the component id.

> Or do you want an API like the one in this PR that just takes org.kde.dolphin/*/* as bundle-ID and fuzzy-matches that?

That. Which is what this wants to do.

qt/pool.cpp
```diff
@@ -175,6 +175,12 @@ QList<Component> Pool::componentsByExtends(const QString &extendedId) const
                                                              qPrintable(extendedId)));
 }
 
+QList<Component> Pool::componentsByBundleId(Bundle::Kind kind, const QString &extendedId) const
+{
+    return cptArrayToQList(as_pool_get_components_by_extends(d->pool,
```
ximion, line 180:
```suggestion
    return cptArrayToQList(as_cache_get_components_by_bundle_id(d->pool,
```

src/as-cache.c
```diff
@@ -1804,6 +1804,34 @@ as_cache_get_components_by_launchable (AsCache *cache, AsLaunchableKind kind, co
 					  error);
 }
 
+/**
+ * as_cache_get_components_by_launchable:
```
ximion, line 1808:
```suggestion
 * as_cache_get_components_by_bundle_id:
```

src/as-cache.c
```diff
@@ -1804,6 +1804,34 @@ as_cache_get_components_by_launchable (AsCache *cache, AsLaunchableKind kind, co
...
+ * @cache: An instance of #AsCache.
+ * @kind: Type of the launchable.
```
ximion, line 1810:
```suggestion
 * @kind: Type of the bundle.
```

src/as-cache.c
```diff
@@ -1804,6 +1804,34 @@ as_cache_get_components_by_launchable (AsCache *cache, AsLaunchableKind kind, co
...
+ * @id: ID of the launchable.
```
ximion, line 1811:
```suggestion
 * @id: ID of the bundle.
```

src/as-cache.c
```diff
@@ -1804,6 +1804,34 @@ as_cache_get_components_by_launchable (AsCache *cache, AsLaunchableKind kind, co
...
+ * @error: A #GError or %NULL.
+ *
+ * Get components which provide a certain launchable.
```
ximion, line 1814:
```suggestion
 * Get components which are provided by a bundle with the given kind and ID.
```

src/as-cache.h
```diff
@@ -4,7 +4,7 @@
  *
  * Licensed under the GNU Lesser General Public License Version 2.1
  *
- * This library is free software: you can redistribute it and/or modify
+ * This library is free software: you can redistribuste it and/or modify
```
ximion, line 7: typo, and probably accidental change

src/as-pool.c
```diff
@@ -2213,6 +2213,33 @@ as_pool_get_components_by_extends (AsPool *pool, const gchar *extended_id)
...
+
+	result = as_cache_get_components_by_bundle_id (priv->cache, kind, bundle_id, &tmp_error);
+	if (result == NULL) {
+		g_warning ("Unable find components offering the bundle in session cache: %s", tmp_error->message);
+		return g_ptr_array_new_with_free_func (g_object_unref);
+	}
+
+	return result;
+}
+
```
ximion, line 2242: this needs a GIR wrapper, like the one below too... But I could add that as a follow-up commit, if you don't want to dive into writing binding code.

src/as-pool.c
```diff
@@ -2213,6 +2213,33 @@ as_pool_get_components_by_extends (AsPool *pool, const gchar *extended_id)
 	return result;
 }
 
+/**
+ * as_pool_get_components_by_bundle_id: (skip)
+ * @pool: An instance of #AsPool.
+ * @kind: The kind of the bundle we are looking for
+ * @bundle_id: The ID, as specified in AsBundle
```
ximion, line 2220:
```suggestion
 * @bundle_id: The ID, as specified in #AsBundle
```

src/as-pool.c
```diff
@@ -2213,6 +2213,33 @@ as_pool_get_components_by_extends (AsPool *pool, const gchar *extended_id)
...
+ *
+ * Find components that provide a bundle with a specific id.
```
ximion, line 2222:
```suggestion
 * Find components that are provided by a bundle with a specific ID.
```

src/as-pool.c
```diff
@@ -2213,6 +2213,33 @@ as_pool_get_components_by_extends (AsPool *pool, const gchar *extended_id)
...
+as_pool_get_components_by_bundle_id (AsPool *pool, AsBundleKind kind, const gchar *bundle_id)
+{
+	AsPoolPrivate *priv = GET_PRIVATE (pool);
+	g_autoptr(GError) tmp_error = NULL;
+	GPtrArray *result = NULL;
+	g_autoptr(GRWLockReaderLocker) locker = g_rw_lock_reader_locker_new (&priv->rw_lock);
+
+	result = as_cache_get_components_by_bundle_id (priv->cache, kind, bundle_id, &tmp_error);
+	if (result == NULL) {
+		g_warning ("Unable find components offering the bundle in session cache: %s", tmp_error->message);
```
ximion, line 2236:
```suggestion
		g_warning ("Unable find components by bundle ID in session cache: %s", tmp_error->message);
```

ximion (changes requested):
Okay :-)
Do you think you'd ever want some "fuzzy matching" for bundle IDs? (I don't think so, but just in case, so we don't need to adjust public API again later).

Also, a bunch of comments on the code ;-)

src/as-pool.c
aleixpol: Please, I'd appreciate it if you could look into this later.

aleixpol:
> Do you think you'd ever want some "fuzzy matching" for bundle IDs? (I don't think so, but just in case, so we don't need to adjust public API again later).

I can see the use-case for wanting to match `org.kde.dolphin/*` and `org.kde.dolphin/x86_64/master` and even `org.kde.dolphin/*/master`.

I don't think we need more flexibility than that, but considering here we are supporting different formats from flatpak, I'm not sure we can come up with a simple API to cover this other than the wildcards.

ximion:
The problem is that I don't think libxmlb supports wildcards to match text strings... It can fuzzy-match strings as described in https://github.com/hughsie/libxmlb/blob/567d261706add40cbe6a5acbd2a3f504b80449de/src/xb-string.c#L141-L147, but doesn't have support for wildcards, AFAIK.
I'll have a look...

qt/pool.cpp
```diff
@@ -175,6 +175,12 @@ QList<Component> Pool::componentsByExtends(const QString &extendedId) const
                                                              qPrintable(extendedId)));
 }
 
+QList<Component> Pool::componentsByBundleId(Bundle::Kind kind, const QString &extendedId) const
+{
+    return cptArrayToQList(as_cache_get_components_by_bundle_id(d->pool,
```
ximion, line 180:
```suggestion
    return cptArrayToQList(as_pool_get_components_by_bundle_id(d->pool,
```

src/as-pool.c
ximion: It's quite trivial, so, no problem :-)

PR #459, opened 2023-01-26
Use 'may' instead of 'my' in as-relation.c

rffontenelle:
It was a typo, I assume.


ximion:
Indeed! Thanks for the patch!

PR #461, opened 2023-01-29
docs: Clarify the locations where catalog icons should be placed

pabloyoyoista:
Rationale: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1574#note_1625290

### DIFF
--- a/docs/xml/catalog-iconcache.xml
+++ b/docs/xml/catalog-iconcache.xml
@@ -20,9 +20,11 @@
 	<section id="spec-iconcache-location">
 		<title>Filesystem locations</title>
 		<para>
-			All icons of type <literal>cached</literal> must be placed in <filename>/usr/share/swcatalog/icons/%{origin}/%{size}/</filename> or <filename>/var/cache/swcatalog/icons/%{origin}/%{size}/</filename>,
+			All icons of type <literal>cached</literal> must be placed in <filename>/usr/share/swcatalog/icons/%{origin}/%{size}/</filename> or <filename>/var/(lib|cache)/swcatalog/icons/%{origin}/%{size}/</filename>,
 			where <literal>origin</literal> is the AppStream data origin defined in the AppStream data file (see <xref linkend="spec-asxml-general"/>), and <literal>size</literal>
 			is <code>64x64</code> or <code>128x128</code> depending on the size of the icon. And icon might be present with different sizes in both directories.
+			When finding cached icons, they must only be searched for in the same location where the accompanying catalog metadata is located.
+			For example, icons for <filename>/var/lib/swcatalog/xml/debian-jessie-main.xml.gz</filename> must only be looked up in <filename>/var/lib/swcatalog/icons/debian-jessie-main/</filename> and not also in locations in <filename>/var/cache/</filename> or elsewhere.
 		</para>
 		<para>
 			For example the cache icon <code>krita.png</code> of a component in a data file with the origin <code>jessie</code> should be stored in


docs/xml/catalog-iconcache.xml
```diff
@@ -20,9 +20,10 @@
 	<section id="spec-iconcache-location">
 		<title>Filesystem locations</title>
 		<para>
-			All icons of type <literal>cached</literal> must be placed in <filename>/usr/share/swcatalog/icons/%{origin}/%{size}/</filename> or <filename>/var/cache/swcatalog/icons/%{origin}/%{size}/</filename>,
+			All icons of type <literal>cached</literal> must be placed in <filename>/usr/share/swcatalog/icons/%{origin}/%{size}/</filename> or <filename>/var/lib/swcatalog/icons/%{origin}/%{size}/</filename>,
```
ximion, line 23: `/var/cache/` is still a valid location, so maybe change this to 
```suggestion
			All icons of type <literal>cached</literal> must be placed in <filename>/usr/share/swcatalog/icons/%{origin}/%{size}/</filename> or <filename>/var/(lib|cache)/swcatalog/icons/%{origin}/%{size}/</filename>,
```

docs/xml/catalog-iconcache.xml
```diff
@@ -20,9 +20,10 @@
...
 			where <literal>origin</literal> is the AppStream data origin defined in the AppStream data file (see <xref linkend="spec-asxml-general"/>), and <literal>size</literal>
 			is <code>64x64</code> or <code>128x128</code> depending on the size of the icon. And icon might be present with different sizes in both directories.
+			The location of the icons in either directory should be done following the same rules as for the XML or YAML catalog metadata.
```
ximion, line 26:
```suggestion
			When finding cached icons, they must only be searched for in the same location where the accompanying catalog metadata is located.
			For example, icons for <filename>/var/lib/swcatalog/xml/debian-jessie-main.xml.gz</filename> must only be looked up in <filename>/var/lib/swcatalog/icons/debian-jessie-main/</filename> and not also in locations in <filename>/var/cache/</filename> or elsewhere.
```

Just a suggestion - I found the suggested sentence a bit confusing, because I don't know what it refers to without the context given by the GNOME Software bug report. Also, and example is always nice to have!

ximion (changes requested):

docs/xml/catalog-iconcache.xml
pabloyoyoista: This is much more readable, thanks!

docs/xml/catalog-iconcache.xml
pabloyoyoista: I would be hesitant to add this change, since the `/var/cache` location is not mentioned anywhere else in the document, including nowhere in the catalog xml or yaml documentation. I would either keep `/var/cache` out of the equation as it currently is, or properly document this currently "edge-case" location. But you are the maintainer, so if you insist, I'll just add the suggestion :)

ximion: You know, `/var/cache` is actually de-facto supported, but maybe there is a chance to phase out support for it... Yet, on the line you changed you changed "cache" to "lib", so I think we need to keep both locations mentioned here for now, until we actually commit to removing one.

pabloyoyoista: Fair enough, I'm always too fast at removing things. Updates as suggested :)

ximion:
Nice, looks good now! Thank you for the patch!

And removal time of deprecated stuff is AppStream 1.0, maybe I'll have a look at this then as well ;-)

PR #463, opened 2023-02-02
spec: Expand documentation for <issue> elements

### LINKED ISSUE #458: The appstream specification is too vague about how to write release notes
https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Releases.html#tag-release-issues

The specification is quite vague about what can be written in the `<description>` and `<issue>` tags for a release.

Long ago I wrote long bulleted lists with links to the issues fixed into `<description>` but then `appstreamcli --validate` complained, it did not want  me to use URLs in the description and I was not able to achieve a nice looking formatted result of the text either. If I remember correctly, list using `ul/li` did not  work in `<description>`.

Therefore, I changed `<description>` to only contain a generic message like `Bugs fixed` and tried to put the detailed information
into the `<issue>` tags, for example like:

```xml
    <release version="2.19.7" date="2022-10-24" urgency="medium">
      <description>
        <p>Bug fixes and enhancements</p>
      </description>
      <url>https://github.com/mike-fabian/ibus-typing-booster/releases/tag/2.19.7</url>
      <issues>
        <issue url="https://github.com/mike-fabian/ibus-typing-booster/issues/395">
          github-issues#395: Avoid inline completion when OSK
          (on-screen-keyboard) is shown
        </issue>
```

But  gnome-software does not display that at all, see:

https://gitlab.gnome.org/GNOME/gnome-software/-/issues/2049#note_1648786

The gnome-software developer Milan Crha and me both think that the specification does not say clearly what exactly to write into the `<issue>` tags. What I did above seems reasonable to me, but when reading the specification I am not sure whether it is OK like that.

Can you please clarify this?

pwithnall:
See the commit messages.

This should hopefully resolve #458, although there may be things I’ve missed or have worded badly.

I also haven’t been able to preview the generated HTML, because I don’t have `daps` installed and from some quick searching around I can’t find it packaged for Fedora. I have run my changes through `xmllint` at least.

Closes #458

### DIFF
--- a/docs/xml/releases-data.xml
+++ b/docs/xml/releases-data.xml
@@ -163,6 +163,7 @@
 				<para>
 					The <literal>issues</literal> tag contains <literal>issue</literal> children defining issues resolved by this release.
 					It is used most commonly to mention <ulink url="https://en.wikipedia.org/wiki/Common_Vulnerabilities_and_Exposures">CVE</ulink> IDs.
+					Software which is interpreting the release notes for the component should present the list of issues separately from the release description. They should not be thought of as a bullet-point list of issues which follow straight on in prose from the <literal>description</literal> element’s value.
 				</para>
 				<para>
 					The value of an <literal>issue</literal> tag must be the bug number, ticket name, or CVE ID and is typically displayed to the user, but may also in case of CVE IDs be read by
--- a/docs/xml/releases-data.xml
+++ b/docs/xml/releases-data.xml
@@ -140,7 +140,12 @@
 			<listitem>
 				<para>
 					A <literal>description</literal> tag contains a brief description of what is new in the release.
-					The <literal>description</literal> tag is structured as described in <xref linkend="tag-description"/>.
+					The intended audience of the description are the users of the component (who are typically not developers), and so the description should mention only the user visible changes in the release.
+					The <literal>description</literal> tag supports child tags as described in <xref linkend="tag-description"/>.
+				</para>
+				<para>
+					It is recommended to not link to issue trackers or bug reports, as these typically make no sense to users.
+					If particular issues need to be highlighted (for example, CVEs fixed in this release), they should be listed in the <literal>issues</literal> tag.
 				</para>
 			</listitem>
 			</varlistentry>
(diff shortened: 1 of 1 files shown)


docs/xml/releases-data.xml
```diff
@@ -167,10 +167,35 @@
...
+					The <literal>issue</literal> tag may have a <literal>type</literal> property, which should have a value of <literal>generic</literal> or <literal>cve</literal>.
+					If the <literal>type</literal> property is missing, a type of <literal>generic</literal> is assumed.
+				</para>
+				<para>
+					It may also have a <literal>url</literal> property, which should be a URL for a details page on the respective issue.
+				</para>
+				<para>
+					If <literal>type</literal> is <literal>cve</literal>, the element’s value must be a CVE ID in the <ulink url="https://www.cve.org/About/Process#cve-id">format defined by MITRE</ulink>.
+					For example, <literal>CVE-2023-12345</literal>. Software consuming the release data of a component should be able to append the element’s value to <literal>https://nvd.nist.gov/vuln/detail/</literal> to get a page of information about the CVE.
+					If a <literal>url</literal> property is given, its value overrides any URL constructed from the CVE identifier.
```
ximion, line 187: I was going to say that this is probably wrong, but checked the implementation, and this is indeed how the implementation behaves. Neat! :-P

ximion (approved):

ximion:
I was wondering whether this is a bit verbose, given that the specification is supposed to be brief and to the point, while we have the [Quickstart Guide](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html) (that should actually be updated to fir the current age...) to be a lot more verbose with a ton more examples.

However, the changes here are still not overly verbose and definitely improve the clarity of the specification. Also, more examples are always very useful!

> I also haven’t been able to preview the generated HTML, because I don’t have `daps` installed and from some quick searching around I can’t find it packaged for Fedora.

Don't worry, the CI does that for you - but you could also just fetch DAPS from Git and even run it from Git. I think there's also an unofficial Fedora package: https://github.com/openSUSE/daps/blob/main/INSTALL.adoc

PR #464, opened 2023-02-07
release: Add sanity checks at beginning of each function

tintou:
Allows to fail with a loud message on what happened before an actual issue.

### DIFF
--- a/src/as-release.c
+++ b/src/as-release.c
@@ -212,6 +212,7 @@ AsReleaseKind
 as_release_get_kind (AsRelease *release)
 {
 	AsReleasePrivate *priv = GET_PRIVATE (release);
+	g_return_val_if_fail (AS_IS_RELEASE (release), AS_RELEASE_KIND_UNKNOWN);
 	return priv->kind;
 }
 
@@ -229,6 +230,7 @@ void
 as_release_set_kind (AsRelease *release, AsReleaseKind kind)
 {
 	AsReleasePrivate *priv = GET_PRIVATE (release);
+	g_return_if_fail (AS_IS_RELEASE (release));
 	priv->kind = kind;
 }
 
@@ -244,6 +246,7 @@ const gchar*
 as_release_get_version (AsRelease *release)
 {
 	AsReleasePrivate *priv = GET_PRIVATE (release);
+	g_return_val_if_fail (AS_IS_RELEASE (release), NULL);
 	return priv->version;
 }
 
@@ -258,6 +261,7 @@ void
 as_release_set_version (AsRelease *release, const gchar *version)
 {
 	AsReleasePrivate *priv = GET_PRIVATE (release);
+	g_return_if_fail (AS_IS_RELEASE (release));
 	as_assign_string_safe (priv->version, version);
 }
 
@@ -273,6 +277,8 @@ as_release_set_version (AsRelease *release, const gchar *version)
 gint
 as_release_vercmp (AsRelease *rel1, AsRelease *rel2)
 {
+	g_return_val_if_fail (AS_IS_RELEASE (rel1), 0);
+	g_return_val_if_fail (AS_IS_RELEASE (rel2), 0);
 	return as_vercmp_simple (as_release_get_version (rel1),
 				 as_release_get_version (rel2));
 }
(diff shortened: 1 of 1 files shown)


src/as-release.c
```diff
@@ -212,6 +212,9 @@ AsReleaseKind
 as_release_get_kind (AsRelease *release)
 {
 	AsReleasePrivate *priv = GET_PRIVATE (release);
+
```
ximion, line 215: nit: For these one-line return/set functions with just one obvious check, I think you can just drop the two empty lines, they just add visual noise.

src/as-release.c
```diff
@@ -273,6 +285,9 @@ as_release_set_version (AsRelease *release, const gchar *version)
 gint
 as_release_vercmp (AsRelease *rel1, AsRelease *rel2)
 {
+	g_return_val_if_fail (AS_IS_RELEASE (rel1), 0);
+	g_return_val_if_fail (AS_IS_RELEASE (rel2), 0);
```
ximion, line 289: Those two checks aren't needed, as `as_release_get_version` already verifies the input.

src/as-release.c
```diff
@@ -706,6 +788,8 @@ as_release_legacy_get_default_artifact (AsRelease *release)
 	AsReleasePrivate *priv = GET_PRIVATE (release);
 	g_autoptr(AsArtifact) artifact = NULL;
 
+	g_assert (AS_IS_RELEASE (release));
```
ximion, line 791: I don't think this check is needed here: It is used in a static and internal function, so the input will already have been verified, and a hard `g_assert` will be compiled out in all builds that aren't debug builds anyway.

src/as-release.c
```diff
@@ -732,6 +816,9 @@ as_release_legacy_get_default_artifact (AsRelease *release)
 guint64
 as_release_get_size (AsRelease *release, AsSizeKind kind)
 {
+
```
ximion, line 819: Spurious newline.

src/as-release.c
```diff
@@ -880,6 +989,10 @@ as_release_load_from_xml (AsRelease *release, AsContext *ctx, xmlNode *node, GEr
 	xmlNode *iter;
 	gchar *prop;
 
+	g_return_val_if_fail (AS_IS_RELEASE (release), FALSE);
+	g_return_val_if_fail (AS_IS_CONTEXT (ctx), FALSE);
+	g_return_val_if_fail (node != NULL, FALSE);
```
ximion, line 994: Not sure if keeping these checks here makes sense - this function as well as `as_release_to_xml_node` and `as_release_load_from_yaml` and `as_release_emit_yaml` are internal functions that are called thousands of times in many occasions and have all their input verified already.
Then again, the extra checks will most likely not slow these functions down by any noticeable amount (but if any of these checks would ever trigger it would be a fatal error anyway).

ximion (changes requested):

ximion:
Thanks! That must have been quite a bit of work to add these - some newer public API has these checks, while they are generally omitted from private and internal API (with some exceptions) as well as older public API (the latter because of laziness, old conventions, and the thought that the checks won't be needed).

src/as-release.c
tintou: It still helps the consumer to know that it is the content that they passed to `vercmp` that is wrong and also to know if it is `rel1` or `rel2`

src/as-release.c
tintou: I feel that it is fine to have the g_assert in static functions to know early when something goes wrong without impacting the final release build, but I'll remove it from here

src/as-release.c
tintou: Indeed, these are internal function, one possibility would be to have these as assert to make sure that nothing is calling them with wrong arguments in the future but that's not for this MR

tintou:
For context, we had a crash in AppCenter on our launch day https://github.com/elementary/appcenter/pull/1993 which could have just been a warning with this MR :slightly_smiling_face:

ximion:
> For context, we had a crash in AppCenter on our launch day [elementary/appcenter#1993](https://github.com/elementary/appcenter/pull/1993) which could have just been a warning with this MR slightly_smiling_face

I assumed there was a story like this behind this PR :-D 
Thanks for the patch!

PR #465, opened 2023-02-17
Make sed invocation more portable

arrowd:
FreeBSD and MacOS variants of `sed` do not accept `-i` parameter without any argument.


ximion (commented):
Looks good, thank you for the patch!

PR #466, opened 2023-02-24
Add API for asking whether the pool is empty

aleixpol:
On certain systems a refresh is necessary in such cases and, in any case, it's useful to know whether queries are pointless.

---

I wonder if it would make sense to be more specific about what we are looking up, since I guess it could be getting confused by fwupd or flatpak...

### DIFF
--- a/src/as-cache.c
+++ b/src/as-cache.c
@@ -1995,3 +1995,16 @@ as_cache_search (AsCache *cache, const gchar * const *terms, gboolean sort, GErr
 
 	return g_steal_pointer (&results);
 }
+
+guint
+as_cache_get_component_count (AsCache* cache)
+{
+	g_autoptr(GError) error = NULL;
+	g_autoptr(GPtrArray) components = as_cache_query_components (cache,
+					  "components/component",
+					  NULL,
+					  0,
+					  FALSE,
+					  &error);
+	return components->len;
+}
--- a/src/as-cache.h
+++ b/src/as-cache.h
@@ -193,6 +193,8 @@ GPtrArray		*as_cache_search (AsCache *cache,
 					  gboolean sort,
 					  GError **error);
 
+guint			as_cache_get_component_count (AsCache *cache);
+
 G_END_DECLS
 
 #endif /* __AS_CACHE_H */
--- a/src/as-pool.c
+++ b/src/as-pool.c
@@ -2888,6 +2888,13 @@ as_pool_set_load_std_data_locations (AsPool *pool, gboolean enabled)
 	}
 }
 
+gboolean
+as_pool_is_empty (AsPool* pool)
+{
+	AsPoolPrivate *priv = GET_PRIVATE (pool);
+	return as_cache_get_component_count (priv->cache) == 0;
+}
+
 /**
  * as_pool_get_os_metadata_cache_age:
  * @pool: An instance of #AsPool.
--- a/tests/test-pool.c
+++ b/tests/test-pool.c
@@ -486,6 +486,8 @@ test_pool_read (void)
 	cpt_s = AS_COMPONENT (g_ptr_array_index (result, 0));
 	g_assert_cmpstr (as_component_get_id (cpt_s), ==, "org.inkscape.Inkscape");
 	g_clear_pointer (&result, g_ptr_array_unref);
+
+	g_assert_false (as_pool_is_empty (dpool));
 }
 
 /**
--- a/qt/pool.cpp
+++ b/qt/pool.cpp
@@ -249,6 +249,11 @@ bool Pool::addComponent(const AppStream::Component& cpt)
     return addComponents(cpts);
 }
 
+bool AppStream::Pool::isEmpty() const
+{
+    return as_pool_is_empty(d->pool);
+}
+
 uint Pool::cacheFlags() const
 {
 #pragma GCC diagnostic push
--- a/qt/pool.h
+++ b/qt/pool.h
@@ -159,6 +159,8 @@ public:
     void overrideCacheLocations(const QString &sysDir,
                                 const QString &userDir);
 
+    bool isEmpty() const;
+
     Q_DECL_DEPRECATED bool addComponent(const AppStream::Component& cpt);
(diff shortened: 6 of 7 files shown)


src/as-cache.c
```diff
@@ -1995,3 +1995,16 @@ as_cache_search (AsCache *cache, const gchar * const *terms, gboolean sort, GErr
 
 	return g_steal_pointer (&results);
 }
+
+gboolean
+as_cache_get_is_empty(AsCache* cache)
+{
+	g_autoptr(GError) error = NULL;
+	g_autoptr(GPtrArray) components = as_cache_query_components (cache,
```
ximion, line 2003: This is *extremely* inefficient, but let's ignore that for now, I can alter the code later to make it more efficient.

src/as-cache.c
```diff
@@ -1995,3 +1995,16 @@ as_cache_search (AsCache *cache, const gchar * const *terms, gboolean sort, GErr
 
 	return g_steal_pointer (&results);
 }
+
+gboolean
+as_cache_get_is_empty(AsCache* cache)
```
ximion, line 2000: Please make this return a component count instead:
```suggestion
guint
as_cache_get_component_count (AsCache* cache)
```

src/as-cache.c
```diff
@@ -1995,3 +1995,16 @@ as_cache_search (AsCache *cache, const gchar * const *terms, gboolean sort, GErr
...
+{
+	g_autoptr(GError) error = NULL;
+	g_autoptr(GPtrArray) components = as_cache_query_components (cache,
+					  "components/component",
+					  NULL,
+					  0,
+					  FALSE,
+					  &error);
+	return components->len > 0;
```
ximion, line 2009:
```suggestion
	return components->len;
```

src/as-pool.c
```diff
@@ -2888,6 +2888,13 @@ as_pool_set_load_std_data_locations (AsPool *pool, gboolean enabled)
 	}
 }
 
+gboolean
+as_pool_get_components_is_empty(AsPool* pool)
```
ximion, line 2892:
```suggestion
as_pool_is_empty (AsPool* pool)
```

src/as-pool.c
```diff
@@ -2888,6 +2888,13 @@ as_pool_set_load_std_data_locations (AsPool *pool, gboolean enabled)
...
+{
+	AsPoolPrivate *priv = GET_PRIVATE (pool);
+	return as_cache_get_is_empty(priv->cache);
```
ximion, line 2895: style and rename
```suggestion
	return as_cache_get_component_count (priv->cache) > 0;
```

tests/test-pool.c
```diff
@@ -266,7 +266,6 @@ test_cache (void)
 					      NULL,
 					      &error);
 	g_assert_no_error (error);
-	g_assert_true (ret);
```
ximion, line 269: Please don't remove check ;-) - we need this assertion.

tests/test-pool.c
```diff
@@ -486,6 +485,8 @@ test_pool_read (void)
 	cpt_s = AS_COMPONENT (g_ptr_array_index (result, 0));
 	g_assert_cmpstr (as_component_get_id (cpt_s), ==, "org.inkscape.Inkscape");
 	g_clear_pointer (&result, g_ptr_array_unref);
+
+	g_assert_false (as_pool_get_components_is_empty(dpool));
```
ximion, line 489:
```suggestion
	g_assert_false (as_pool_is_empty (dpool));
```

ximion (changes requested):
Quite a few changes needed.
AsCache should expose a more low-level API, so returning a component-count makes a lot of sense here, as we will probably want to have that value returned by the pool anyway as well (so a software center can also refresh if the amount of components looks suspiciously low). but the `is_empty` function can be built upon that.
Also, so `_get_` prefix is needed for `_is_*`-style functions.

The way emptyness is determined is super inefficient here, but if you want to I could fix that up for you in a follow-up commit once this PR is merged (at the moment we deserialize all data and then throw it away just to determine the amount of stuff in the cache, so we at least need to stop doing that. But there's even more efficient methods to do determine the amount of cache entries).

src/as-cache.c
aleixpol: Yes, that's what I thought. I imagine it will need changing libxmlb.

src/as-pool.c
```diff
@@ -2888,6 +2888,13 @@ as_pool_set_load_std_data_locations (AsPool *pool, gboolean enabled)
 	}
 }
 
+gboolean
+as_pool_is_empty (AsPool* pool)
+{
+	AsPoolPrivate *priv = GET_PRIVATE (pool);
+	return as_cache_get_component_count (priv->cache) > 0;
```
ximion, line 2895:
```suggestion
	return as_cache_get_component_count (priv->cache) == 0;
```

src/as-cache.h
```diff
@@ -193,6 +193,8 @@ GPtrArray		*as_cache_search (AsCache *cache,
 					  gboolean sort,
 					  GError **error);
 
+guint			as_cache_get_component_count(AsCache *cache);
```
ximion, line 196:
```suggestion
guint			as_cache_get_component_count (AsCache *cache);
```

tests/test-pool.c
```diff
@@ -486,6 +486,8 @@ test_pool_read (void)
 	cpt_s = AS_COMPONENT (g_ptr_array_index (result, 0));
 	g_assert_cmpstr (as_component_get_id (cpt_s), ==, "org.inkscape.Inkscape");
 	g_clear_pointer (&result, g_ptr_array_unref);
+
+	g_assert_false (as_pool_is_empty(dpool));
```
ximion, line 490:
```suggestion
	g_assert_false (as_pool_is_empty (dpool));
```

ximion (changes requested):

ximion (approved):
Looks good now, I will optimize it a bit post-merge :-)

PR #474, opened 2023-03-17
Fix agreement example

JakobDev:
The doc says, that the property must be named `id`, but the example uses `type`.

btw:
Is agreement actually used? anywhere?


ximion:
Thank you for the patch!

PR #475, opened 2023-03-31
trivial: Add a missing space in the <url> spec

wjt:


ximion:
Everyone indeed loves a one-byte patch ;-)
Thanks!

PR #477, opened 2023-04-11
enums: Bring back AS_FORMAT_STYLE_COLLECTION into its enum

tintou:
It is breaking the introspection API by removing its namespace. It is also likely to break the enum gtype stability.


ximion:
Fair. There was a bug in GIR once which made it fail when using `enum = enum` in C, but clearly that's fixed given that the CI is green.
Thanks for the patch!

PR #482, opened 2023-04-12
Check if Release Description is inside description tag

JakobDev:
Take a look at flathub/website#1336 for the reason why this exists

### DIFF
--- a/src/as-validator-issue-tag.h
+++ b/src/as-validator-issue-tag.h
@@ -870,6 +870,11 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("The release timestamp is invalid."),
 	},
 
+	{ "release-description-outside-tag",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("The release description must be put inside a description tag"),
+	},
+
 	{ "artifact-type-invalid",
 	  AS_ISSUE_SEVERITY_ERROR,
 	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
--- a/src/as-validator.c
+++ b/src/as-validator.c
@@ -2090,8 +2090,10 @@ as_validator_check_release (AsValidator *validator, xmlNode *node, AsFormatStyle
 		}
 
 		/* checks if the description is put outside a description tag */
-		if (g_strcmp0 (node_name, "p") == 0 || g_strcmp0 (node_name, "ol") == 0 || g_strcmp0 (node_name, "ul") == 0 || g_strcmp0 (node_name, "li") == 0)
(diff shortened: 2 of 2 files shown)


src/as-validator-issue-tag.h
```diff
@@ -870,6 +870,11 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	  N_("The release timestamp is invalid."),
 	},
 
+	{ "release-description-outside-tag",
+	  AS_ISSUE_SEVERITY_ERROR,
+	  N_("The release description must be put inside a description tag"),
```
ximion, line 875:
```suggestion
	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
	  N_("The release description must be put inside a `description` tag"),
```

Translators might translate this to something odd otherwise ;-)

src/as-validator.c
```diff
@@ -2089,6 +2089,12 @@ as_validator_check_release (AsValidator *validator, xmlNode *node, AsFormatStyle
 			continue;
 		}
 
+		/* checks if the description is put outside a description tag */
+		if (g_strcmp0 (node_name, "p") == 0 || g_strcmp0 (node_name, "ol") == 0 || g_strcmp0 (node_name, "ul") == 0 || g_strcmp0 (node_name, "li") == 0) {
```
ximion, line 2093:
```suggestion
		if (as_str_equal0 (node_name, "p") || as_str_equal0 (node_name, "ol") || as_str_equal0 (node_name, "ul") || as_str_equal0 (node_name, "li")) {
```
Just for nicer readability :-)

ximion (changes requested):
Just a few minor things :-)

JakobDev:
> Previously something like the linked metadata should already have failed validation due to the bad tags in a `release` node...

No. It just gives a Warning, but not an error, which is fine, as there are more tags might be added in the Future, so the Validation will not break on older Versions. But I don't think something like `<p>` will be added, so we can give a Error here.

src/as-validator.c
ximion: Historical reasons - `as_str_equal0` is a later addition. Likely a Coccinelle script could be written to auto-adjust, but it's not that important.
For new code though, using the helper is nice ;-)

ximion:
Warnings will also fail validations right now ;-)

ximion:
Thanks for the patch! - I'll also cherry-pick it to the 0.16.x branch.

JakobDev:
> Warnings will also fail validations right now ;-)

I have taken a look at it again: It is a Info, not a Warning. That's why it doesn't fail.

ximion:
Oof, yeah, we needed this change then :-)

PR #484, opened 2023-04-13
Fix crash in asc_l10n_search_translations_qt

arrowd:


ximion:
This does look odd to me - "We can't change value of the pointer that is marked `g_autofree`." is also a wrong assumption, you absolutely can do that!
We just can't have it point to uninitialized memory. 
After `g_strstr_len (locale, -1, "/");`, `tmp` points to the first occurrence of `/` (should probably be last occurrence, actually). Then we set locale to that position, advancing it by one. This should at worst yield an empty string in `locale`.

So, can you elaborate a bit more on the patch, and maybe create a testcase for the bug? (Check out `test_compose_locale_stats` in `test-compose.c` for some existing tests).

arrowd:
> This does look odd to me - "We can't change value of the pointer that is marked g_autofree." is also a wrong assumption, you absolutely can do that!

Huh? Isn't `g_free()` just boil down to `free()` call? I'm more than sure that you have to pass `free()` the same pointer returned by `malloc()`. How may freeing `g_autoptr` even work if we pass it not the beginning of the data but some location inside?

ximion:
Okay, it was apparently way too late yesterday and I was confused, sorry for that! I didn't even see that your ternary expression included the + 1, eww...

The only thing `g_free` does over `free` is deal with NULL pointers.

Again, really sorry for the confusion! The patch actually looks fine as-is then!

arrowd:
Great, thanks! Can you please also take a look at #485 ? It is important to us be sure that AppStream will keep working on FreeBSD.

ximion:
On why this worked before btw, I can only guess that either an allocator was used that is fine with this (dlmalloc I think can do that), or that the undefined behavior wasn't bad enough. Either way, it was a bug.

I'll have a look at that PR later, but I guess I'll need an account with Cirrus for that? There's some code that can spawn FreeBSD VMs on GitHub actions, that may be easier to maintain in the long run...

pabloyoyoista:
Just for the record, we have seen this recurrently in alpine linux with muslc, and this fixes the issue, thanks both!

PR #485, opened 2023-04-13
Add FreeBSD CI

Zoid-dr:
Add FreeBSD CI


ximion:
Thank you for the patch! Apparently the newer one from @arrowd indeed obsoletes this one :-)

PR #492, opened 2023-04-21
Introduce release name

aleasto:


ximion:
Can you give examples of what such a "release name" would be? By being implemented as an XML attribute, there would be no way to translate it. Is that really a good idea?

aleasto:
It's a minor detail, but the Fedora Rawhide release name is "rawhide". When showing the release to the user, we would prefer showing it capitalized instead, as "Rawhide".

Not very important, but maybe you can see it useful in other cases.

Conan-Kudo:
For Ubuntu users, this would make sense too. For example, Ubuntu 23.10 is `mantic` but the user-facing version is `Mantic Minotaur`.

ximion:
This would definitely have to be a `name` element in a `release` though, and not an XML attribute... (for future-proofing and possibly localization).

PR #496, opened 2023-05-11
Fix tweaking of appstream.pc when building as subproject

dsd:
Building libadwaita via jhbuild with meson-0.64.0,

  meson.build:227:4: ERROR: Problem encountered: Unable to fix up the .pc file:
  sed: can't read /home/jhbuild/.cache/jhbuild/build/libadwaita/subprojects/appstream/meson-private/appstream.pc: No such file or directory

The generated file is found at /home/jhbuild/.cache/jhbuild/build/libadwaita/meson-private/appstream.pc

Use the global build root to find this file and fix the build failure.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -221,7 +221,7 @@ sed_prog = find_program('sed')
 pc_fixup = run_command(sed_prog,
                        '-i ""',
                        '/^Requires.private\|^Libs.private/ d',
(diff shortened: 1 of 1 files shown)


ximion:
Makes sense! This hack sucks in general, annoyingly it's still the only way to avoid other projects having to pull in development libraries that AppStream needs but they do not...

Thank you for the patch!

PR #497, opened 2023-05-28
Add support for heading in app description 

CarlSchwan:
This is useful for adding some organization inside a long description. h1 is not allowed since this should be the title of the application, and this should not be handled by the app itself.

E.g. this would make sense for the description in https://apps.kde.org/zanshin/

### DIFF
--- a/docs/xml/metainfo-component.xml
+++ b/docs/xml/metainfo-component.xml
@@ -247,6 +247,9 @@
 			<para>Unordered list (<literal>ul</literal>), with list items (<literal>li</literal>)</para>
 			</listitem>
 			<listitem>
+			<para>Heading (<literal>heading</literal>)</para>
+			</listitem>
+			<listitem>
 			<para>
 				Within paragraphs and list items, emphasis (<literal>em</literal>) and inline code (<literal>code</literal>) text styles are supported.
 				The emphasis is commonly rendered in italic, while inline code is shown in a monospaced font.
--- a/src/as-utils.c
+++ b/src/as-utils.c
@@ -259,6 +259,28 @@ as_description_markup_convert (const gchar *markup, AsMarkupKind to_kind, GError
 			} else {
 				g_string_append_printf (str, "%s\n", clean_text);
 			}
+		} else if ((g_strcmp0 ((gchar*) iter->name, "heading") == 0)) {
+			g_autofree gchar *clean_text = NULL;
+			g_autofree gchar *text_content = as_xml_get_node_value_raw (iter);
+			g_auto(GStrv) spl = NULL;
+
+			/* Apparently the element is empty, which is odd. But we better add it instead
+			 * of completely ignoring it. */
+			if (text_content == NULL)
+				text_content = g_strdup ("");
+
+			/* remove extra whitespaces and linebreaks */
+			clean_text = as_sanitize_text_spaces (text_content);
+
+			if (str->len > 0)
+				g_string_append (str, "\n");
+
+			spl = as_markup_strsplit_words (clean_text, 100);
+			if (spl != NULL) {
+				for (guint i = 0; spl[i] != NULL; i++) {
+					g_string_append_printf (str, "## %s\n", clean_text);
+				}
+			}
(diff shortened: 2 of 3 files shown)


ximion:
Thank you for providing an example, those are always useful!
The description texts are intended to be very basic on purpose, so people do not go too crazy and make insanely long articles about their software. The intent is to give users the fastest possible way to see if they want to install the software or not.

That said, I do see why people would like some way of sectioning text, but I do not like using `h2` very much. First of all, the description text in AppStream are *explicitly* intended to not be HTML, adding `h2` would make people believe more that they can use HTML, and much worse, it will naturally lead them to believe that there was a `h1` too - that's a pretty huge trap we would set just for a slightly convenient implementation.

So, what I would suggest is adding just one organizational element, `section`, for now. That one is not HTML, quite clear in what it does, and can be styled by software centers in every way they want. Adding subsections will likely lead to people not making terse descriptions anymore, so I'm not sure if we need those (we could always add them later, I do not like adding any feature in the hope that they will become useful in future).

What do you think? @hughsie do you have any feedback on this issue, given that we avoided adding a feature like this for quite a while... (the 1.0 release would be the time to add such a thing though, if we want it).
I actually see this being quite useful in release descriptions as well, actually.

hughsie:
I don't think we want h2/h3 -- it's just confusing to define. Section makes 100% sense.

JakobDev:
The problem I see with a `section` is, that it is not HTML. I know, that the Description is intended to not be HTML, but all valid Tags for a Description are also valid HTML tags that do the same, so in fact, it is currently a subset of HTML. Sites who Display the AppStream Description relay on this and display just us the Description Code as it is. Adding a new tag that is not HTML would break this.

CarlSchwan:
> The problem I see with a `section` is, that it is not HTML. I know, that the Description is intended to not be HTML, but all valid Tags for a Description are also valid HTML tags that do the same, so in fact, it is currently a subset of HTML. Sites who Display the AppStream Description relay on this and display just us the Description Code as it is. Adding a new tag that is not HTML would break this.

yeah this is also why I choose h2 and h3 since both Discover and apps.kde.org won't need any changes to support it

ximion:
Relying on undefined behavior is not okay though, and adding a hack to the specification just so other tools *might* not need changes (I know some that adding any new tag here would break) is also not okay.

Websites can add CSS styles to `section` elements, while we can provide a to_html helper function in libappstream to convert description markup to HTML (just like we already have for Markdown). That would actually also allow us to add more tags in future with a bit less pain than currently.

CarlSchwan:
`<section>` is already an HTML tag which a different semantic https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section . So if `<h2>` is not acceptable, I would prefer using `<heading>`, which doesn't already mean something.

Similarly using `<strong>` would be quite incorrect since it's used "for content that is of "strong importance," including things of great seriousness or urgency (such as warnings)."  (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong) and so using that for heading can confuse a screen reader

I checked a few other app stores and apple app store, android play store and the windows store don't support headings and generally have extremely short app descriptions. Android and Windows don't even seems to support bold or italic and apps developers add stars to mark stuff as more important. 

Steam, GoG and Epics do support headings and generally have a way bigger app description and inline images.

itch.io  

Maybe this is

CarlSchwan:
friendly ping :)

ximion:
I would like to revisit this, as we can likely safely introduce this (it won't be usable for a while, unfortunately, but once it's in we can at least use it *in future* at some point).
@CarlSchwan are you still interested in this? Is `heading` still seen as a good name for this markup element? (I think that is the best name)

We could have it in 1.2.x, so it can be used in a few years after people stopped avoiding it (it won't render with the current parsers, so headings will simply be missing).

ximion:
This is now implemented as new markup. It will take a while until you will be able to use it liberally though, the ecosystem is slow.

CarlSchwan:
Nice thanks!

PR #503, opened 2023-06-19
cli: Add "get-latest-version" util command

hadess:
This gets the latest version specified in an appdata file, which,
combined with utilities such as check-news.sh[1], can be used to check
whether the date specified in the appdata's release information is
correct.

See https://gitlab.gnome.org/GNOME/evince/-/issues/1950

[1]: https://gitlab.gnome.org/GNOME/totem/-/blob/master/check-news.sh


tools/appstreamcli.c
```diff
@@ -1340,9 +1360,14 @@ as_client_run (char **argv, int argc)
 			as_client_run_check_license);
 	ascli_add_cmd (commands,
 			2, "is-satisfied", NULL, "FILE|COMPONENT-ID",
-			/* TRANSLATORS: `appstreamcli `check-license` command description. */
+			/* TRANSLATORS: `appstreamcli `is-satisfied` command description. */
 			_("Check if requirements of a component (via its ID or MetaInfo file) are satisfied on this system."),
 			as_client_run_is_satisfied);
+	ascli_add_cmd (commands,
+		       2, "get-latest-version", NULL, "FILE",
```
ximion, line 1367: I wonder if this should be called `metainfo-latest-version` or something, to make clear we're not querying the latest AppStream version... But the confusion potential is probably not *that* high...

tools/appstreamcli.c
```diff
@@ -1340,9 +1360,14 @@ as_client_run (char **argv, int argc)
...
+		       /* TRANSLATORS: `appstreamcli get-latest-version command description. */
+		       _("Get the latest version from the AppData file"),
```
ximion, line 1369:
```suggestion
		       _("Get the latest version from the MetaInfo file"),
```
(the file format is called MetaInfo, AppData is an ancient name from an early draft)

tools/ascli-actions-mdata.c
```diff
@@ -739,3 +739,69 @@ ascli_check_is_satisfied (const gchar *fname_or_cid, const gchar *cachepath, gbo
...
+
+	metad = as_metadata_new ();
+	as_metadata_set_locale (metad, "ALL");
+
+	if ((g_str_has_suffix (fname, ".yml.gz")) ||
+	    (g_str_has_suffix (fname, ".yaml.gz")) ||
+	    (g_str_has_suffix (fname, ".yml")) ||
+	    (g_str_has_suffix (fname, ".yaml"))) {
+		/* if we have YAML, we also automatically assume a catalog style */
+		as_metadata_set_format_style (metad, AS_FORMAT_STYLE_CATALOG);
```
ximion, line 771: Does it make any sense to read this for `CATALOG` data? Since catalog files contain many components, you would currently just get a random version, so I don't think that makes sense.
I think it would be better if this code would just error-out in that case and limit itself to metainfo files.
The added benefit of that would be that this code would be much smaller then too ;-)

tools/ascli-actions-mdata.c
```diff
@@ -739,3 +739,69 @@ ascli_check_is_satisfied (const gchar *fname_or_cid, const gchar *cachepath, gbo
...
+	cpt = g_object_ref (as_metadata_get_component (metad));
+
+	if (as_component_get_releases (cpt)->len > 0) {
+		GPtrArray *releases = as_component_get_releases (cpt);
+		AsRelease *release_newest = NULL;
+
+		for (guint i = 0; i < releases->len; i++) {
+			AsRelease *release_tmp = g_ptr_array_index (releases, i);
+			if (release_newest == NULL ||
+			    as_release_get_timestamp (release_tmp) > as_release_get_timestamp (release_newest))
```
ximion, line 796: We may not always have a timestamp (e.g. for development versions / unreleased ones).
Since the `release` entries must be sorted highest to lowest, we could cheat here and just pick the first release entry, or we could go the extra mile of comparing version numbers if one of the releases has no timestamp...

Is it common that there are sorting errors with `release` elements? (`appstreamcli validate` should throw an error for that).

tools/ascli-actions-mdata.c
```diff
@@ -739,3 +739,69 @@ ascli_check_is_satisfied (const gchar *fname_or_cid, const gchar *cachepath, gbo
...
+	as_metadata_parse_file (metad,
+				infile,
+				AS_FORMAT_KIND_UNKNOWN,
+				&error);
+	if (error != NULL) {
+		g_printerr ("%s\n", error->message);
+		return 3;
+	}
+
+	cpt = g_object_ref (as_metadata_get_component (metad));
```
ximion, line 787: nitpick: You don't need to reference the AsComponent here, as it is held for the duration of AsMetadata's life. That said, it also doesn't really hurt either.

ximion (changes requested):

PR #504, opened 2023-06-26
Add support for specifying the platform of a given screenshot

### LINKED ISSUE #390: platform dependent screenshots
I'd like to suggest to add the ability to tag (<tags>?) screenshots or otherwise constrain them to a given platform. For example Krita, the drawing application, may look different in a plasma environment than in a gnome environment than in a solus environment. To give a more native/integrated feel the authors may choose to offer platform dependent screenshots.

This would incidentally also allow appstream data to feed into the windows and apple stores, which arguably deserve their own set of native screenshots ;)

CarlSchwan:
Fix #390

Questions is if we should reuse the triplet for the platform of the artifact which aren't that relevant or specify a bunch of platforms (gnome, gnome-mobile, android, plasma desktop, plasma mobile, windows, macos, ...) as standard values and allow the app developer to also put other values for e.g. some obscure os like haiku.

### DIFF
--- a/docs/xml/catalog-xmldata.xml
+++ b/docs/xml/catalog-xmldata.xml
@@ -431,6 +431,9 @@
 						on the image shown. The tag is translatable.
 					</para>
 					<para>
+						Optionally, a screenshot can contain a <code><![CDATA[<platform>]]></code> tag, describing which platform the screenshot is for.
+					</para>
+					<para>
 						For <code><![CDATA[<screenshot>]]></code> tags that contain <literal>video</literal> elements, a catalog metadata generator may impose any restrictions to them,
 						including completely removing them from the output, imposing filesize limits, etc.
 						Upstream metainfo files should not rely on the videos being present and must always have a static screenhot for the software component as well.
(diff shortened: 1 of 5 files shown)


ximion:
I'm unsure about the semantics here, as a `platform` in AppStream is a triplet of `arch-kernel-userspace`, and does not encode whether the screenshot was done under KDE or GNOME. I don't think we will want architecture-specific screenshots, tbh.

ingokloecker:
> I'm unsure about the semantics here, as a `platform` in AppStream is a triplet of `arch-kernel-userspace`, and does not encode whether the screenshot was done under KDE or GNOME. I don't think we will want architecture-specific screenshots, tbh.

That's correct. Primarily, we want to add screenshots for Windows (for use in Microsoft Store and other Windows app stores), for Android (for Android stores), for macOS (for Apple Store), etc. Those screenshots shouldn't be shown by Linux app stores (which probably can already be achieved by leaving out the `type="default"` attribute). See #514 for our ugly workaround.

Maybe `os` would be a better term than the overloaded `platform` term. It would somewhat exclude the different Linux desktops, but I don't really see a use case for screenshots of Gnome apps with KDE Plasma window decoration and vice-versa in AppStream.

danirabbit:
> I don't really see a use case for screenshots of Gnome apps with KDE Plasma window decoration and vice-versa in AppStream.

There are some apps that have different behaviors based on the current desktop. For example Epiphany will load Granite styles while in Pantheon. I think there's value for apps which try to support multiple fdo platforms

CarlSchwan:
Not needed anymore

PR #509, opened 2023-07-26
qt: Allow loading Pool asynchronously

kbroulik:
Maps as_pool_load_async.

### DIFF
--- a/qt/pool.cpp
+++ b/qt/pool.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2016 Matthias Klumpp <matthias@tenstral.net>
+ * Copyright (C) 2023 Kai Uwe Broulik <kde@broulik.de>
  *
  * Licensed under the GNU Lesser General Public License Version 2.1
  *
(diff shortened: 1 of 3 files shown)


qt/pool.cpp
```diff
@@ -65,9 +68,24 @@ pool_changed_cb (AsPool *cpool, AppStream::Pool *qpool)
...
 }
 
+static void
+pool_ready_async_cb (AsPool *cpool, GAsyncResult *result, gpointer user_data)
+{
+    auto *d = static_cast<PoolPrivate *>(user_data);
+    g_autoptr(GError) error = NULL;
+
+    if (as_pool_load_finish (cpool, result, &error)) {
+        Q_EMIT d->q->loaded();
```
ximion, line 78: Hmm... I wonder whether instead of having a `loaded` and `failedToLoad` signal, we should just have one `loadFinished(success)` signal instead that takes a boolean to indicate whether loading was a success or wasn't. What do you think?

kbroulik: Merged them into a `loadFinished(bool)` signal.

ximion:
Nice, thank you!

PR #510, opened 2023-07-26
 Port to Windows(2) 

elmarco:
supersedes #501


elmarco:
There is "compose" left now... can be addressed in a follow-up.

taozuhong:
Dependency xmlb from subproject subprojects/libxmlb found: YES 0.3.12
Run-time dependency libsystemd found: NO (tried pkgconfig and cmake)
Dependency gobject-introspection-1.0 found: YES 1.76.1 (cached)
Library stemmer found: YES

**meson.build:166:23: ERROR: Include dir /usr/include does not exist.**

A full log can be found at D:/Github/appstream/build/meson-logs/meson-log.txt

data/meson.build
```diff
@@ -7,6 +7,11 @@ install_data('its/metainfo.its',
 install_data('its/metainfo.loc',
               install_dir: join_paths(get_option('datadir'), 'gettext', 'its'))
 
+# Do not rely on an exe wrapper for rel-info, use the system one in that case
+if meson.is_cross_build()
+    ascli_exe = find_program('appstreamcli')
```
ximion, line 12: This would fetch `appstreamcli` from the environment - is that really needed? Reason is that, since ascli comes from AppStream, it may not exist in the environment, or may even exist as an older version. We often use newer features in AppStream itself first, so it is quite important that AppStream validates its own files with the tools it has just built, and doesn't fail due to using older tools from the environment.

meson.build
```diff
@@ -35,9 +35,6 @@ conf.set_quoted('LOCALSTATEDIR',
 conf.set_quoted('LIBEXECDIR',
                 join_paths(get_option('prefix'),
                            get_option('libexecdir')))
-conf.set_quoted('DATADIR',
```
ximion, line 38: This is a super weird error? Do you have any idea what causes this? (Just in case we need to reintroduce `DATADIR` in future)

elmarco: it get expanded to
```c
typedef IDataObject *LPDATAOBJECT;
typedef enum tagDATADIR {
    DATADIR_GET = 1,
    DATADIR_SET = 2
} 
 "/usr/x86_64-w64-mingw32/sys-root/mingw/include/objidl.h"
```

because objidl.h has:
```c
10666:typedef enum tagDATADIR {
10667:    DATADIR_GET = 1,
10668:    DATADIR_SET = 2
10669:} DATADIR;
```

data/meson.build
elmarco: It's only needed for cross-compilation. Otherwise we would need to build a native version too - which is a lot more complicated, and should proabably be done as a sub-build somehow... tricky

ximion: > Iow, it's a lot simpler to install a native version before doing a cross-compilation. If you worry about versions, we could check the native appstream.pc version..

That would be nice to or, or maybe at least emit a warning in the log file, so debugging any issues becomes a bit easier :-)

elmarco: added

ximion (approved):
This looks good! Thank you for the work!
What is this used for on Windows? Would it make sense to add Windows CI to AppStream as well?

elmarco:
> This looks good! Thank you for the work! What is this used for on Windows? Would it make sense to add Windows CI to AppStream as well?

It is used by libadwaita. Having CI is always nice :) but I am not very familiar with github CI.

meson.build
```diff
@@ -164,9 +160,10 @@ if get_option ('gir')
     dependency('gobject-introspection-1.0', version: '>=1.56')
 endif
 
-stemmer_inc_dirs = include_directories(['/usr/include'])
+stemmer_inc_dirs = include_directories()
 if get_option('stemming')
     stemmer_lib = cc.find_library('stemmer', required: true)
+    stemmer_inc_dirs = include_directories(['/usr/include'])
```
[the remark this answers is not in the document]
elmarco: yeah, I haven't looked at stemmer much.. they don't support cross-compilation anyway.. But yeah, they should provide a .pc file...

ximion:
> It is used by libadwaita.

Interesting! I didn't know that :-)

> Having CI is always nice :) but I am not very familiar with github CI.

I'm not very familiar with Windows or setting it up for Windows, but maybe I can have a look later. In any case, I think this is good to go!
This will be part of the upcoming 1.0 release of AppStream, which still has ways to go (unfortunately I couldn't work on it as much as I wanted so far), but that also means this will receive a lot of testing :-)

PR #514, opened 2023-08-14
Allow translation of custom values

ingokloecker:
Add support for translated custom values. The validator checks that there is only one value tag for each key and language. The XML reader ignores translated custom values. You have to use another reader if you need the translated custom values.

To be able to use the AppStream data as source for our submissions to the Microsoft Store we have added custom values that need to be translated to the AppStream data of NeoChat (and soon other KDE applications). More specifically we have added custom values for Windows-specific screenshots including (translated) captions. Our tooling does already support translated custom values. To be able to continue using the validator we would like the validator to allow translated custom values. The code of the XML reader simply ignores translated custom values for backward compatibility.

Appstream data of NeoChat with translated custom values:
https://invent.kde.org/network/neochat/-/blob/master/org.kde.neochat.appdata.xml#L269

### DIFF
--- a/po/sv.po
+++ b/po/sv.po
@@ -27,7 +27,7 @@ msgstr ""
 "Project-Id-Version: appstream-glib\n"
 "Report-Msgid-Bugs-To: appstream@lists.freedesktop.org\n"
 "POT-Creation-Date: 2023-09-30 00:32+0200\n"
-"PO-Revision-Date: 2023-10-01 00:22+0000\n"
+"PO-Revision-Date: 2023-10-02 11:00+0000\n"
 "Last-Translator: Anders Jonsson <anders.jonsson@norsjovallen.se>\n"
 "Language-Team: Swedish <https://hosted.weblate.org/projects/appstream/translations/sv/>\n"
 "Language: sv\n"
@@ -1483,7 +1483,7 @@ msgstr ""
 
 #: src/as-spdx-data.h:29
 msgid "Amazon Digital Services License"
-msgstr ""
+msgstr "Amazon Digital Services-licensen"
 
 #: src/as-spdx-data.h:30
 msgid "Academic Free License v1.1"
@@ -1531,11 +1531,11 @@ msgstr "GNU Affero General Public License v3.0 eller senare"
 
 #: src/as-spdx-data.h:41
 msgid "AMD's plpa_map.c License"
-msgstr ""
+msgstr "AMD:s plpa_map.c-licens"
 
 #: src/as-spdx-data.h:42
 msgid "Apple MIT License"
-msgstr ""
+msgstr "Apples MIT-licens"
 
 #: src/as-spdx-data.h:43
 msgid "Academy of Motion Picture Arts and Sciences BSD"
@@ -1555,7 +1555,7 @@ msgstr ""
 
 #: src/as-spdx-data.h:47
 msgid "Adaptive Public License 1.0"
-msgstr ""
+msgstr "Adaptive Public License 1.0"
 
 #: src/as-spdx-data.h:48
 msgid "Apple Public Source License 1.0"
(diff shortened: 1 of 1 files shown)


ximion:
Hmm, so, while I do want to find a solution here, I do see some issues:

* I can't accept any implementation that is incomplete and spits out different data from what is read in. If we validate a file as "valid AppStream", users of the library and AppStream tooling must expect the tooling to handle said data, and not silently strip it from the file e.g. when adding release information or processing metainfo files into catalog data.
* Since all tags are "value" tags, we can't create good ITS rules for this to make this translatable. We can't even make these things optionally translatable without causing issues.
* Putting lists into a key-value table that does not support them (`KDE::windows_store::screenshots::1`, `KDE::windows_store::screenshots::2`, ...) is super ugly...

I do get why you want this data in a MetaInfo file though, it makes thematic sense to keep it there. We can not, however, as previously discussed, add support for random features of other stores that we have no influence over to AppStream - that would make the project unmaintainable in the long run.

So, I see two paths forward, one easier, one more complicated:

#### Solution A
It looks like you are not using AppStream's tooling to read this data anyway (since it simply doesn't work and doesn't validate this stuff right now), so apparently there is something parsing the raw XML. It looks like you are also not using features like `release-to-metainfo` where release data is added to MetaInfo files from a YAML or NEWS file. If that is the case, why not use `x-` prefixed properties for this? Have a `x-microsoft-store` root node, and throw all the stuff you want below it. The validator will ignore anything prefixed with `x-`. That way, you get a way cleaner implementation of this, and it will even pass the validator checks!
The downside of that approach is that none of AppStream's tooling will ever read that data, and it will be stripped if the XML is processed with it.

#### Solution B
We could potentially expand AppStream to just allow *any* node construct below `custom` (maybe limiting its depth a little), and read & export the node tree for consumers to read. That way, AppStream's tools could take care of that data as well and just serialize/deserialize it with everything else, integrating it very well. It would also have all the benefits of Solution A. The downside of this approach is a bit of overhead when reading data, and providing people with a very convenient escape hatch that might prevent them from adding potential new features to AppStream properly (but I don't think that's too likely), and it may be slightly more difficult to filter custom data in generator tools. It would also be a breaking change, but the point to make such a change would be right now, in the AppStream 1.0 cycle.

What do you think?

I am also curious about the opinions of @hughsie and @pwithnall on option B, since they may have some further insights. Adding this just for KDE is a lot of work, so if there was more demand for this extensive feature, that would make it easier to justify spending time on it (of course, I also gladly accept complete PRs ;-) - but this one would be tricky to add properly).

hughsie:
> KDE::windows_store::screenshots::1::caption

This is pretty terrible -- we already have a translated screenshot caption. It's basically unmaintainable.

ingokloecker:
Thanks for the feedback! Yes, the workaround for custom value tags just supporting text content is super ugly. I'm very much in favor of allowing any node construct (within reason) below `custom`. Or maybe a second tag next to `value` which would allow for any node construct below it. Maybe I should have just asked for something like this in the first place when I realized that custom values are just text. On the other hand, the ugly example is probably more convincing.

I might be able to spend some time on this as part of my paid work for KDE e.V., but I'm not experienced in glib.

ingokloecker:
@hughsie: Any opinion on option B proposed by @ximion? Or maybe a comment on #504 which would also solve our use case and make this PR obsolete.

hughsie:
From a reading-the-data point of view, any node-centric system like libxmlb is going to skip over any sub-node of <custom> in one pointer read and a uint64 add -- so no concerns about performance there. I'd have more issues with including the custom sub-nodes in AppStream data (rather than just in metainfoinfo.xml files) due to size worries -- so we'd have to say "preserve the sub-nodes when modifying the metainfo.xml files, but disregard them when building metadata" for instance.

ximion:
The `appstream-generator` will include nothing in the `custom` section in generated custom data. Anything that is passed through needs to be explicitly whitelisted. Filtering things will be a lot harder though if it is not just a key-value store but a whole node tree...

PR #516, opened 2023-08-21
bundle: add linglong as a new type of Bundle

BLumia:
This is a preparation for adding AppStream metadata support for linglong packages. Thus 'linglong' can be used as a valid bundle type in its AppStream catalog metadata.

cc @black-desk

### DIFF
--- a/docs/xml/catalog-xmldata.xml
+++ b/docs/xml/catalog-xmldata.xml
@@ -597,6 +597,7 @@
 							<listitem><para><code>snap</code> - For <ulink url="https://snapcraft.io/">Snappy</ulink> snap bundles.</para></listitem>
 							<listitem><para><code>tarball</code> - For plain and possibly compressed tarballs.</para></listitem>
 							<listitem><para><code>cabinet</code> - For cabinet firmware deployments.</para></listitem>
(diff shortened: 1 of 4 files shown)


ximion:
What is a linglong package? Do you have a link?
If it is like a DEB/RPM package to compose a Linux distribution, then this would not qualify as a bundling system but instead as a distro packaging system and you could just use the `package` tag instead (but would need to add support for this packaging format to appstream-generator).

BLumia:
It's https://linglong.dev/en/

It's a bundling system similar to flatpak, currently maintained by the developers from deepin (and also used by its commercial distribution, UOS).

Let me know if you need any additional information. Thanks!

BLumia:
Sorry for the ping, do you need any additional information to be provided in order to get this merged? Thanks!

BLumia:
I'm not part of the Linglong team but I've heard they are planning to introduce granular and transparent permission control and other stuff related to improve OCI compatibility. Right now it's not that different with Flatpak but come with some uncommon architectures support like loongarch. Hopefully they will eventually improve Linglong's official website to provide more information about Linglong itself :)

PR #520, opened 2023-09-01
Add FreeBSD CI using Cirrus CI.

arrowd:
Testing is omitted for the moment because AsMonitor tests fail due to FreeBSD-specific bug in glib: https://gitlab.gnome.org/GNOME/glib/-/issues/576

I might also miss some dependencies, so the commit might need amending after the repository enables Cirrus CI.

Obsoletes https://github.com/ximion/appstream/pull/485 PR

### DIFF
--- a/.cirrus-ci.yml
+++ b/.cirrus-ci.yml
@@ -1,0 +1,17 @@
+task:
+  freebsd_instance:
+    matrix:
+      - image_family: freebsd-13-2-snap
+  env:
+    # /usr/ports/Mk/Uses/localbase.mk localbase:ldflags
+    LOCALBASE: /usr/local
+    CFLAGS: -isystem $LOCALBASE/include
+    CPPFLAGS: $CFLAGS
+    CXXFLAGS: $CFLAGS
+    LDFLAGS: -L$LOCALBASE/lib
+  deps_script:
+    - sed -i.bak 's/quarterly/latest/' /etc/pkg/FreeBSD.conf
+    - env ASSUME_ALWAYS_YES=yes pkg update -f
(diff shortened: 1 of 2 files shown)


ximion:
To make this a bit more maintainable, would it be possible to make Cirrus call `tests/ci/run-build.sh` and use different build options of *BSD? Or does FreeBSD not support `/etc/os-release`, so adding support for it would be more tedious in that file? (it would be nice to configure all CI builds in one place...)

arrowd:
Sure, it seems we can reuse that.

.cirrus-ci.yml
```diff
@@ -0,0 +1,18 @@
...
+    CFLAGS: -isystem $LOCALBASE/include
+    CPPFLAGS: $CFLAGS
+    CXXFLAGS: $CFLAGS
+    LDFLAGS: -L$LOCALBASE/lib
+  deps_script:
+    - sed -i.bak 's/quarterly/latest/' /etc/pkg/FreeBSD.conf
+    - env ASSUME_ALWAYS_YES=yes pkg update -f
+    - env ASSUME_ALWAYS_YES=yes pkg install -y snowballstemmer gperf itstool gtk-doc qt6-tools
+  build_script:
+    - sed -i '' "s/'-i'/'-i \"\"'/" src/meson.build
```
ximion, line 17: Is this still needed?

ximion (changes requested):

arrowd:
I see you changed this line 2 weeks ago. I guess, no it shouldn't be needed now.

ximion:
Good! I guess this will not actually build before I merge it, so let's merge it now and hopefully it will be fine (I will probably need you to address any non-obvious issues, as I am less familiar with FreeBSD).

ximion:
Looks like that build environment has no C compiler. Can you please fix that?

PR #522, opened 2023-09-07
data: Add ‘endless’ to the list of desktops

pwithnall:
It’s used to identify Endless OS desktops, which use a modified version of GNOME.

It’s always used alongside GNOME (i.e. `DesktopNames=endless:GNOME`), but is useful as a separate identifier to allow distinguishing between modified Endless OS desktops and stock GNOME.

It’s defined in `/usr/share/wayland-sessions/endless-wayland.desktop` on Endless OS systems. That file is provided by the `gnome-session` package whose Debian packaging data is modified during packaging. (i.e. There’s no git repository I can point you to which defines the `endless` desktop ID.)

Defining the desktop allows metainfo files which specify, for example, `<compulsory_for_desktop>endless</compulsory_for_desktop>` to pass validation.


pwithnall:
I’ve also proposed it [here](https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/73) for the XDG menu spec.

ximion:
> I’ve also proposed it [here](https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/73) for the XDG menu spec.

I was about to ask that wile reading the original description :laughing: 
The non-capitalized spelling is intentional, right? (this value will be case-sensitive)

pwithnall:
> The non-capitalized spelling is intentional, right? (this value will be case-sensitive)

Yes, it is intentional, as `endless` is what we use already in various bits of software. See https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/73#note_2075562.

I’ve just asked internally, however, about whether we should consider changing the capitalisation now (before it’s standardised) to neaten things up. Please hold!

ximion:
> Please hold!

Okay, just tell me when it's ready :-) This patch is otherwise fine obviously, so I can merge it as soon as you're fine with it :-)

ximion:
> But it could probably happen if the lowercase version is too unpalatable for the spec.

FTR, it's fine with me, it's just that this is the only entry that doesn't start with a capital letter, so it looks a bit like a mistake in the list ^^ 
But if it's intentional, spec-wise I see no issue (as long as we don't change it once it has been in a release).

dbnicholson:
> > Please hold!
> 
> Okay, just tell me when it's ready :-) This patch is otherwise fine obviously, so I can merge it as soon as you're fine with it :-)

We're switching it to capitalized `Endless`, so I think we'll want to change the patch.

pwithnall:
I’ve updated the patch to use the new capitalisation. From my point of view it’s ready to merge. Thanks!

ximion:
Nice, I have no objections :-)
Thanks for the patch!

PR #525, opened 2023-09-08
Some random meson improvements

tristan957:
Adding overrides is the biggest thing.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2,7 +2,7 @@ project('AppStream', 'c',
   meson_version: '>=0.62',
   default_options: ['c_std=c11', 'cpp_std=gnu++17'],
 
-  license: 'LGPL-2.1+',
+  license: 'LGPL-2.1-or-later',
   version: '1.0.0',
 )
 
--- a/src/meson.build
+++ b/src/meson.build
@@ -204,6 +204,8 @@ appstream_dep = declare_dependency(
     include_directories: [appstream_lib_inc]
 )
 
+meson.override_dependency('appstream', appstream_dep)
+
 pkgc.generate(
     appstream_lib,
     version: as_version,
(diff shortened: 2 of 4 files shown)


ximion:
Enabling fatal warnings may not have been such a good idea ;-)
(I explicitly want to add more strict warning flags when building in maintainer mode, but build with relaxed options in production, so GCC updates do not cause sudden mass build failures when a new warning was added. Meson unfortunately doesn't allow this scheme and goes for an "always or never" approach, which I don't think is good at all. If there's a way to achieve the same result with Meson though, that would be fine with me.)

tristan957:
Should I fix this `-Werror` warning?

tristan957:
I don't see a reason to ever let Meson warn, honestly. Compiler warnings like "-Wdecl-after-statement" can get annoying when doing random things however.

ximion:
> Should I fix this `-Werror` warning?

Only if the current behavior is retained: All warnings enabled + `-Werror` when in maintainer mode, no `-Werror` when building without maintainer mode.
Right now, I see no other way to do that than to remove the added `--fatal-meson-warnings`, but if you know a better solution, I'd be happy for that!

ximion:
> Compiler warnings like "-Wdecl-after-statement" can get annoying when doing random things however.

If you're not in maintainer mode, that's just a warning and not an error. It will become one if you try to upstream the changes though.

tristan957:
I'm just giving an example of how CIs should enforce -Werror, not build systems.

ximion:
I do not understand what you mean. The CI does enforce -Werror, it is not enabled by default.

ximion:
If you drop the `--fatal-meson-warnings` change for now (until Meson is changed to allow to conditionally enable warnings using a native method), this change is good to go and should also pass CI.

ximion:
I've merged all the changes except for the CI-breaking one. Thank you for the patch!

Issue #519, opened 2023-08-30
as_get_physical_memory_total  missing macOS support


taozuhong:
```
static gulong
as_get_physical_memory_total (void)
{
#if defined(__linux__)
	struct sysinfo si = { 0 };
	sysinfo (&si);
	if (si.mem_unit > 0)
		return (si.totalram * si.mem_unit) / MB_IN_BYTES;
	return 0;
#elif defined(__FreeBSD__)
	unsigned long physmem;
	sysctl ((int[]){ CTL_HW, HW_PHYSMEM }, 2, &physmem, &(size_t){ sizeof (physmem) }, NULL, 0);
	return physmem / MB_IN_BYTES;
#elif defined(G_OS_WIN32)
	MEMORYSTATUSEX statex;

	statex.dwLength = sizeof (statex);
	GlobalMemoryStatusEx (&statex);
	return statex.ullTotalPhys / (1024 * 1024);
#else
#error "Implementation of as_get_physical_memory_total() missing for this OS."
#endif
}
```

ximion:
Yes, patches welcome! ;-)

alex-tee:
is something like this what's missing? https://stackoverflow.com/questions/583736/determine-physical-mem-size-programmatically-on-osx

PR #526, opened 2023-09-09
fix as-system-info compilation on mac

alex-tee:
It looks like it's the same logic as FreeBSD (Mac is based on BSD after all): https://stackoverflow.com/questions/583736/determine-physical-mem-size-programmatically-on-osx

(hopefully) fixes https://github.com/ximion/appstream/issues/519

### DIFF
--- a/src/as-system-info.c
+++ b/src/as-system-info.c
@@ -49,8 +49,15 @@
 
 #if defined(__linux__)
 #include <sys/sysinfo.h>
-#elif defined(__FreeBSD__)
+#elif defined(__FreeBSD__) || defined(__APPLE__)
 #include <sys/types.h>
+# ifdef __APPLE__
+/* for some reason these need to be defined otherwise compilation fails
+ * when including sys/sysctl.h */
+typedef unsigned int u_int;
+typedef unsigned short u_short;
+typedef unsigned char u_char;
+# endif
 #include <sys/sysctl.h>
 #endif
 #ifdef HAVE_SYSTEMD
@@ -412,7 +419,7 @@ as_get_physical_memory_total (void)
 	if (si.mem_unit > 0)
 		return (si.totalram * si.mem_unit) / MB_IN_BYTES;
 	return 0;
-#elif defined(__FreeBSD__)
+#elif defined(__FreeBSD__) || defined(__APPLE__)
 	unsigned long physmem;
 	sysctl ((int[]){ CTL_HW, HW_PHYSMEM }, 2, &physmem, &(size_t){ sizeof (physmem) }, NULL, 0);
 	return physmem / MB_IN_BYTES;
--- a/src/as-system-info.c
+++ b/src/as-system-info.c
@@ -51,13 +51,13 @@
 #include <sys/sysinfo.h>
 #elif defined(__FreeBSD__) || defined(__APPLE__)
 #include <sys/types.h>
-# ifdef __APPLE__
+#ifdef __APPLE__
 /* for some reason these need to be defined otherwise compilation fails
  * when including sys/sysctl.h */
 typedef unsigned int u_int;
 typedef unsigned short u_short;
 typedef unsigned char u_char;
-# endif
+#endif
 #include <sys/sysctl.h>
 #endif
 #ifdef HAVE_SYSTEMD


ximion:
Have you tested this on macOS? Merging untested code is very iffy...

alex-tee:
I'm building my project in GitHub actions - please wait... will report back when I get this to work. For some reason Mac doesn't define uint/ushort/uchar even though it uses those in its own headers... very strange

alex-tee:
the current version of the PR passes compilation on mac and my software runs

ximion:
Does it also return a sensible value? This code doesn't have unit tests, for obvious reasons (mocking this would bypass the code we would want to test ^^).

src/as-system-info.c
```diff
@@ -49,8 +49,15 @@
 
 #if defined(__linux__)
 #include <sys/sysinfo.h>
-#elif defined(__FreeBSD__)
+#elif defined(__FreeBSD__) || defined(__APPLE__)
 #include <sys/types.h>
+# ifdef __APPLE__
```
ximion, line 54:
```suggestion
#ifdef __APPLE__
```
nit: no space after `#` please

src/as-system-info.c
```diff
@@ -49,8 +49,15 @@
...
+/* for some reason these need to be defined otherwise compilation fails
+ * when including sys/sysctl.h */
+typedef unsigned int u_int;
+typedef unsigned short u_short;
+typedef unsigned char u_char;
+# endif
```
ximion, line 60:
```suggestion
#endif
```

alex-tee:
> Does it also return a sensible value? This code doesn't have unit tests, for obvious reasons (mocking this would bypass the code we would want to test ^^).

I have no idea, sorry, I just needed this to build libadwaita. I don't have time to look too much into this, but my app with libadwaita seems to run fine

ximion:
> I have no idea, sorry, I just needed this to build libadwaita. I don't have time to look too much into this, but my app with libadwaita seems to run fine

Let's test this, it will be really easy to do!
Please create an XML file like this on your Mac:
```xml
<?xml version="1.0"?>
<component type="generic">
  <id>org.example.test</id>
  <name>TestApp</name>

  <requires>
    <memory>406900</memory>
  </requires>
</component>
``` 
I doubt your machine has 397 GiB RAM, in case it does, just increase the memory value a bit ;-)

Then use the freshly compiled `appstreamcli` and run this command:
`appstreamcli is-satisfied /path/to/your/dummy.xml` (replace the path to the just created XML file).
You should then see output like this:
```console
Relation check for: */*/*/org.example.test/*

Requirements:
 ✘ This software requires >= 397.36 GiB of memory, but this system has 30.48 GiB.
Recommendations:
 • No recommended items are set for this software.
Supported:
 • No supported items are set for this software.
``` 
If it gives the right amount of memory here, then everything is fine and we can merge this :-)

ximion:
Or even easier, just run `appstreamcli sysinfo` :-D

dehesselle:
Mac mini M1, macOS Monterey 12.7

Output from `LANG=C ./appstreamcli sysinfo`
```bash
Operating System Details:

(appstreamcli:73406): GLib-CRITICAL **: 19:02:03.203: g_utf8_strdown: assertion 'str != NULL' failed

(appstreamcli:73406): XbValueBindings-CRITICAL **: 19:02:03.203: xb_value_bindings_bind_str: assertion 'str != NULL' failed
? Unable to find operating system component '(null)'!

Kernel:
Name: Darwin
Version: 21.6.0

Hardware:
Physical Memory: 3577 MiB (3.49 GiB)

** (appstreamcli:73406): WARNING **: 19:02:03.203: Error while searching for modalias entries in /sys/devices: No such file or directory

User Input Controls:

** (appstreamcli:73406): WARNING **: 19:02:03.203: Unable to read input info: Unable to look for input device: AppStream was built without systemd-udevd support.
 ? console: yes
```
Output from `LANG=C ./appstreamcli is-satisfied test.xml`
```bash
Relation check for: */*/*/org.example.test/*

Requirements:
 ? This software requires >= 397.36 GiB of memory, but this system has 7.49 GiB.
Recommendations:
 ? No recommended items are set for this software.
Supported:
 ? No supported items are set for this software.
```

ximion:
That looks sane, I guess your Mac has 8GB of RAM?

dehesselle:
Oh I should have paid more attention. While I did notice that `sysinfo` shows a different value than the `is-satisfied` test and just chose to not bother, they're actually both wrong. The machine in question has 16 GB memory.

ximion:
So, apparently on macOS you are supposed to use `HW_MEMSIZE` and not `HW_PHYSMEM`, because the latter will give you nonsense values.
Apparently on macOS, physmem may be 32bit or may just give you "usable physical memory", instead of actual physical memory. Hurray... :-P

dehesselle:
That works! Outputs are now

```text
Operating System Details:

(appstreamcli:77778): GLib-CRITICAL **: 01:23:21.076: g_utf8_strdown: assertion 'str != NULL' failed

(appstreamcli:77778): XbValueBindings-CRITICAL **: 01:23:21.077: xb_value_bindings_bind_str: assertion 'str != NULL' failed
? Unable to find operating system component '(null)'!

Kernel:
Name: Darwin
Version: 21.6.0

Hardware:
Physical Memory: 16384 MiB (16.00 GiB)

** (appstreamcli:77778): WARNING **: 01:23:21.077: Error while searching for modalias entries in /sys/devices: No such file or directory

User Input Controls:

** (appstreamcli:77778): WARNING **: 01:23:21.077: Unable to read input info: Unable to look for input device: AppStream was built without systemd-udevd support.
 ? console: yes
```
and
```text
Relation check for: */*/*/org.example.test/*

Requirements:
 ? This software requires >= 397.36 GiB of memory, but this system has 16.00 GiB.
Recommendations:
 ? No recommended items are set for this software.
Supported:
 ? No supported items are set for this software.
```

ximion:
Perfect! @alex-tee, can you update the patch? We will need a new macOS-specific codepath for this and can't reuse the FreeBSD one.
I'd be happy to merge this patch then :-)

ximion:
Perfect! Thank you @dehesselle for testing this! @alex-tee, can you update the patch? We will need a new macOS-specific codepath for this and can't reuse the FreeBSD one.
I'd be happy to merge this patch then :-)

ximion (changes requested):
Perfect! Thank you @dehesselle for testing this! @alex-tee, can you update the patch? We will need a new macOS-specific codepath for this and can't reuse the FreeBSD one.
I'd be happy to merge this patch then :-)

ximion:
Ping?

ximion:
@dehesselle Can you please check if the current changes in Git master compile and work on macOS? Potentially parts of this PR are still needed...

dehesselle:
This is the first time I'm building `master`. While the `as-system-info` problem of this PR is fixed and appstream builds fine, I cannot build `libadwaita` with it:

```
../../../usr/src/libadwaita-1.4.0/src/adw-about-window.c:2007:14: error: implicit declaration of function 'as_component_get_releases' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
  releases = as_component_get_releases (component);
             ^
../../../usr/src/libadwaita-1.4.0/src/adw-about-window.c:2007:14: note: did you mean 'as_component_set_releases'?
../../../include/appstream/as-component.h:289:13: note: 'as_component_set_releases' declared here
void           as_component_set_releases (AsComponent *cpt, AsReleaseList *releases);
               ^
../../../usr/src/libadwaita-1.4.0/src/adw-about-window.c:2007:12: warning: incompatible integer to pointer conversion assigning to 'GPtrArray *' (aka 'struct _GPtrArray *') from 'int' [-Wint-conversion]
  releases = as_component_get_releases (component);
           ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../../usr/src/libadwaita-1.4.0/src/adw-about-window.c:2041:20: error: implicit declaration of function 'as_component_get_developer_name' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
  developer_name = as_component_get_developer_name (component);
                   ^
../../../usr/src/libadwaita-1.4.0/src/adw-about-window.c:2041:20: note: did you mean 'as_component_get_developer'?
../../../include/appstream/as-component.h:248:15: note: 'as_component_get_developer' declared here
AsDeveloper  *as_component_get_developer (AsComponent *cpt);
              ^
../../../usr/src/libadwaita-1.4.0/src/adw-about-window.c:2041:18: warning: incompatible integer to pointer conversion assigning to 'const char *' from 'int' [-Wint-conversion]
  developer_name = as_component_get_developer_name (component);
                 ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 warnings and 2 errors generated.
```

I'm assuming that `libadwaita` as a downstream project will have to adopt to these changes later, but in the meantime I'd like for this commit to be backported to 0.16.x so I can use a (persumably) 0.16.4 in my builds at some point.

ximion:
Right. This should be backportable, but Adwaita can also surely be made to build with any AppStream version. Thanks for confirming this works! :-)

Issue #511, opened 2023-08-11
finding appstreamqt finds qt6 when building with qt5


hsitter:
I have an application that does
```
find_package(AppStreamQt REQUIRED)
```
which now seems to find the qt6 version of appstreamqt despite me having written that line for qt5 and the application is also still building for qt5.
Presumably the cmake config name needs versioning or something?

ximion:
This is kind of intentional: AppStream 1.0 is Qt6-only and bumped it's SONAME, so the Qt5 and Qt6 version are coinstallable.
The development headers for both are not however (well... not without some tricks) - so you can either develop with the old 0.16.x series of AppStream and Qt5, or the new (unreleased) AppStream 1.0 series using Qt6 only.

ximion:
Is cmake smart enough to figure this out with this change? Or do I need to rename the package name to `AppStreamQt5` as well, to make the Qt version obvious to cmake?

PR #529, opened 2023-09-14
add to the qt dual build

carlosdem:
hi.  from a packaging point of view it makes much more sense to leave qt5 as and make qt6 self-titled.  it's much easier to alter unreleased code than to try alter existing packaging expectations.  with that in mind the naming scheme has been reversed and unfortunately cmake was not smart enough as you feared in https://github.com/ximion/appstream/issues/511 so that has been resolved in the qt6 cmake files.  kde neon unstable is carrying this patch at present and also patches porting all the qt6 usage of appstream in invent.kde.  all the legacy qt5 usage seems happy enough after some version requirement bumps.  thanks for the coinstallability it makes life much easier and means i can delete my half-finished attempt in my own fork.  cheers


ximion:
No, I don't want to do that. The API for 1.0 is broken anyway, so this is not a simple Qt6 rebuild, existing Qt5 code needs to be adapted anyway as well.
(A lot of API has been removed, renamed or its behavior has changed)
Therefore, I'd rather make this explicit. And since the Qt5 library will be phased out eventually, the new Qt6 library gets the generic name.

carlosdem:
hi.  thanks for the feedback.

fair enough.  completely understand your reasons.  

the problematic part is that https://github.com/ximion/appstream/issues/511 is not fixed as such.  agreed, the development headers for both will now be installable as long as all usage of  <=0.16.3 is removed entirely from the system.   0.16.* and 1.0.* are not co-installable.  even if you strip 0.16.3 down to just a qt5 build it will still conflict with qt6 because of libAppStreamQt.so .. the solib version's for both toolkits have indeed been bumped but they share a common name for a base which the versioned solib's symbolically link to.

effectively the dual build fixes nothing in the short term during the transition period nor in the long term for things that will never get ported.

we've been trying to carry 2 versions, 1.0 for kf6 and plasma6 but also 0.16.* for qt5 and kf5 libs for many months now.  the *5 ecosystem is definitely not yet legacy and will not be for quite a while to come.

as libAppStreamQt.so will continue to exist as the base for two different libraries from 2 different toolkits, i'm not sure what to suggest other than release and give an ultimatum for all to adapt or not be supported.

PR #532, opened 2023-10-01
allow Qt5 to build with std optional and ==int

carlosdem:
toInt was introduced in Qt6.2 and qt5 seems to need the explicit inclusion of "optional" in a few places.


qt/component-box.cpp
```diff
@@ -30,7 +30,7 @@ class AppStream::ComponentBoxData : public QSharedData
 public:
     ComponentBoxData(ComponentBox::Flags flags)
     {
-        m_cbox = as_component_box_new(static_cast<AsComponentBoxFlags>(flags.toInt()));
+        m_cbox = as_component_box_new(static_cast<AsComponentBoxFlags>(flags==int()));
```
ximion, line 33: I have never seen that `flags==int()` syntax - aren't we converting a nonsense bool to `AsComponentBoxFlags` here with that?

AFAIK casting QFlags to int should also have the same effect as explicitly calling toInt, so maybe that's a solution (but if there's some magic going on here that I don't know, can you explain it? I searched the Qt documentation and didn't find anything mentioning that this works).

ximion (changes requested):

ximion:
Thank you for the patch, but I think this one has been superseded by https://github.com/ximion/appstream/commit/33fbce7408a31f56c84d476d8eb1ab025461f68b and its subsequent patches :-)

PR #533, opened 2023-10-02
Add GUI environment IDs for macOS and Microsoft Windows

ingokloecker:
Thanks for adding support for screenshot environments!

This adds IDs for macOS and Microsoft Windows.


ximion:
That looks fine, thank you! :-)

PR #534, opened 2023-10-04
qt: Improve ComponentBox

aleixpol:
* Adds iterators so that we don't need to create a full QList every time that we get a result. We can iterate it little by little and only keep in memory what we need.
* Adds operator+= and erase() so the ComponentBox can be operated on.

### DIFF
--- a/qt/component-box.cpp
+++ b/qt/component-box.cpp
@@ -137,6 +137,18 @@ void ComponentBox::sortByScore()
     as_component_box_sort_by_score(d->m_cbox);
 }
 
+void ComponentBox::operator+=(const ComponentBox& other)
+{
+    for (uint i = 0; i < as_component_box_len(other.d->m_cbox); i++) {
+        g_autoptr(GError) error = nullptr;
+        AsComponent *c = as_component_box_index(other.d->m_cbox, i);
+        as_component_box_add(d->m_cbox, c, &error);
+        if (error) {
+            qWarning() << "error adding component" << error->message;
+        }
+    }
+}
+
 ComponentBox::iterator ComponentBox::erase(iterator it)
 {
     as_component_box_remove_at(it.data->d->m_cbox, it.index);
--- a/qt/component-box.cpp
+++ b/qt/component-box.cpp
@@ -136,3 +136,9 @@ void ComponentBox::sortByScore()
 {
     as_component_box_sort_by_score(d->m_cbox);
 }
+
+ComponentBox::iterator ComponentBox::erase(iterator it)
+{
+    as_component_box_remove_at(it.data->d->m_cbox, it.index);
+    return it;
+}
--- a/qt/component-box.cpp
+++ b/qt/component-box.cpp
@@ -61,6 +61,13 @@ public:
     AsComponentBox *m_cbox;
 };
 
+Component ComponentBox::iterator::operator*() const
+{
+    AsComponent *cpt = as_component_box_index(data->d->m_cbox, index);
+    Q_ASSERT(cpt != nullptr);
+    return Component(cpt);
+}
+
 ComponentBox::ComponentBox(const ComponentBox &other)
     : d(other.d)
 {
--- a/qt/component-box.h
+++ b/qt/component-box.h
@@ -40,6 +40,41 @@ class APPSTREAMQT_EXPORT ComponentBox
     Q_GADGET
 
 public:
+    class iterator {
+    public:
+        iterator(const iterator& it)
+            : index(it.index)
+            , data(it.data)
+        {}
+        bool operator==(iterator other) const { return index == other.index && data == other.data; }
+        bool operator!=(iterator other) const { return index != other.index || data != other.data; }
+        iterator& operator++() { index++; return *this; }
+
+        Component operator*() const;
+        std::optional<Component> operator->() const {
+            return data->indexSafe(index);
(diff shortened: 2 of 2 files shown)


aleixpol:
Discover's counterpart: https://invent.kde.org/plasma/discover/-/merge_requests/650

qt/component-box.h
```diff
@@ -40,6 +40,39 @@ class APPSTREAMQT_EXPORT ComponentBox
...
 
 public:
+    class iterator {
+    public:
+        bool operator==(iterator other) const { return index == other.index && data == other.data; }
+        bool operator!=(iterator other) const { return index != other.index || data != other.data; }
+        iterator& operator++() { index++; return *this; }
+
+        Component operator*() const {
+            return *data->indexSafe(index);
```
ximion, line 50: This will return `NULL` / `std::nullopt` if `index` is out of bounds. What happens then?
If `index` can never become out of bounds, we can potentially give the iterator access to an unsafe accessor method in future to save the "index within bounds" check that `indexSafe` does.

ximion (changes requested):

qt/component-box.h
aleixpol: I think an unsafe call is what we want here, I figured it can be done once everything is in place.

qt/component-box.cpp
```diff
@@ -125,3 +125,20 @@ void ComponentBox::sortByScore()
 {
     as_component_box_sort_by_score(d->m_cbox);
 }
+
+AppStream::ComponentBox::iterator AppStream::ComponentBox::erase(iterator it)
+{
+    GPtrArray *array = as_component_box_array(it.data->d->m_cbox);
+    g_ptr_array_remove_index(array, it.index);
```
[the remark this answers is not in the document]
aleixpol: It makes Discover's patch _much_ cleaner as we often need to filter them. I'll look into what you say...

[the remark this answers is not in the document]
ximion: Removing stuff from a box isn't currently allowed... If it's really needed, I could add support for it though.

qt/component-box.h
ximion: The unsafe query should probably be a private/protected member of the component box then, so it's not called by accident :-)

qt/component-box.cpp
```diff
@@ -125,3 +125,18 @@ void ComponentBox::sortByScore()
...
+    return it;
+}
+
+void ComponentBox::operator+=(const ComponentBox& other)
+{
+    for (uint i = 0; i < as_component_box_len(other.d->m_cbox); i++) {
+        AsComponent *c = as_component_box_index(other.d->m_cbox, i);
+        as_component_box_add(d->m_cbox, c, nullptr);
```
ximion, line 140: Is there any way that we could propagate an error here? Currently, components may silently not be added... (I don't see a sane way to do that, so this might just be as good as it gets...)

qt/component-box.cpp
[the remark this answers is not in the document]
aleixpol: If it's not allowed, I need to add a bunch of workarounds in discover (see MR there). I added it in the glib API, please see if this is what you had in mind.

qt/component-box.h
aleixpol: Moved it to the cpp file, no API leakage.

qt/component-box.cpp
aleixpol: I'll add a warning, then we can decide to extend the API if we see it's a real problem.

qt/component-box.cpp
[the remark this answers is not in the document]
ximion: There was a few issues with regards to memory management and the way the internal index is handled, sorry for the reply delay and only seeing this now!
I've implemented a remove_at function that should work well in pretty much all scenarios that I can currently thin k of :-)

ximion:
Can you rebase and use the existing new `as_component_box_remove_at` function? :)

ximion:
Ping ;-)

qt/component-box.cpp
```diff
@@ -61,6 +61,13 @@ class AppStream::ComponentBoxData : public QSharedData
     AsComponentBox *m_cbox;
 };
 
+Component ComponentBox::iterator::operator*() const
+{
+    AsComponent *cpt = as_component_box_index(data->d->m_cbox, index);
+    Q_ASSERT(cpt != nullptr);
```
ximion, line 67: This can not be NULL... But that check also doesn't hurt.

qt/component-box.cpp
```diff
@@ -129,3 +136,22 @@ void ComponentBox::sortByScore()
...
+        if (error) {
+            qWarning() << "error adding component" << error->message;
+        }
+    }
+}
+
+ComponentBox::iterator ComponentBox::erase(iterator it)
+{
+    GPtrArray *array = as_component_box_array(it.data->d->m_cbox);
+    g_ptr_array_remove_index(array, it.index);
```
ximion, line 155: You *must not* do this, it will break the component box (the returned box is constant, but unfortunately C/C++ will let you do this even though it shouldn't be done). Use `as_component_box_remove_at` instead.

ximion (changes requested):

ximion:
This looks good now, thank you for the patch!

PR #536, opened 2023-10-15
Add NetBSD support.

0-wiz-0:


ximion:
This one is trivial to add, but is there any way we could just enable this for all the *BSDs? Or are they too different so doing this does not make sense?

PR #539, opened 2023-10-18
Add ignore-empty option to validate

JakobDev:
If you run `appstreamcli validate` with no files given, it will exit with a `You need to specify at least one file to validate!` error. With this new option it no longer throws a error.

This is useful for scenarios like
```
find -name "*.metainfo.xml" | xargs appstreamcli validate --explain
```


JakobDev:
Ping @ximion

PR #541, opened 2023-10-24
Support building for Windows with clang-cl

fanc999-1:
Hi,

This attempts to add build support for Windows using clang-cl (the flavor of clang that attempts to mimick Visual Studio and uses the Microsoft SDKs, meaning `_MSC_VER` will also be defined), so that the build libraries and headers can also be better consumed by items built using Visual Studio.  This is done to restore support to build libadwaita against the Microsoft tools/SDK.

This mainly consists of:

* Making the code compile with the Windows SDK by only including headers when they are really available.
* Using compiler directives to export symbols from shared builds via defining macros.  Also, for the private symbols, define macros suitable for clang-cl for visibility.  clang-cl (and Visual Studio) build DLLs with hidden visibility by default.
* Improve dep search using CMake names where needed also, as CMake build files for Visual Studio builds often do not produce pkg-config files for us.  Also avoid hard-coding paths to find the libstemmer/snowball headers.  Also do not enable systemd on Windows, since it is (and is unlikely ever) not available for Windows at all

With blessings, thank you!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -18,6 +18,9 @@ as_minor_version = varr[1]
 as_micro_version = varr[2]
 as_data_installdir = get_option('prefix') / get_option('datadir') / 'appstream'
 
+# Are we building for Windows?
+is_windows = host_machine.system() == 'windows'
+
 #
 # Configure files
 #
@@ -78,7 +81,7 @@ if get_option('maintainer')
     add_project_arguments(maintainer_c_args, language: 'cpp')
 endif
 
-if get_option('static-analysis') and host_machine.system() != 'windows'
+if get_option('static-analysis') and not is_windows
   if cc.get_id() != 'gcc'
     error('You need to compile with GCC to run the static analyzer!')
   endif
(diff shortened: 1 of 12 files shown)


fanc999-1:
> > Using compiler directives to export symbols from shared builds via defining macros.
> 
> Can this be done differently? I really don't like this, and it adds a ton of unnecessary noise to the code (it also doesn't look like other projects like GLib need this?).
> 
> The other changes look okay from a quick glance, thank you a lot for this work! Having CI for this will be very helpful! :-)

I'll try to see whether there are any better ways for this--GLib/Pango/GTK do have something like this though, however, via the `GLIB_AVAILABLE*` macros that decorate symbols, in GLib for instance, and on Visual Studio and clang-cl they will be defined as `__decspec(dllexport)` during the build.

Thanks for the suggestions though.

ximion:
It's a bit ugly, but I guess we need to accept that :-)
Why is the CI failing though?

ximion:
Any updates on this? This PR needs a rebase for quite a while now...

ximion:
Closing due to inactivity for more than 2 years.

PR #543, opened 2023-10-26
qt: Bring back Utils::currentDistroComponentId

Sodivad:
It looks like it whas accidentally removed in
8cc4db43c386f0f8201fc7da9e14fa9ee9314998.

### DIFF
--- a/qt/utils.cpp
+++ b/qt/utils.cpp
@@ -22,6 +22,11 @@
 #include "appstream.h"
 #include "chelpers.h"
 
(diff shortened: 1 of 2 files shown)


ximion:
No, use `SystemInfo::currentDistroComponentId()` or `SystemInfo::osCid()` instead.
This function should never have been part of utils (and isn't in the C library, so the Qt one is now in sync with the C library).

Sodivad:
Thanks for the hints, I will port discover to those. I just saw there was still a function that was wrapped

PR #546, opened 2023-10-31
Unbreak FreeBSD CI.

arrowd:


arrowd:
It is green now, whew.

I should've probably set up Cirrus for myself first and prepare the PR on my local account. Sorry for the CI trashing I introduced while fixing it.

ximion:
At least it works now :-) Thank you for fixing it so quickly!

arrowd:
Thanks for merging this. Please ping me if you get any FreeBSD CI breakages.

Issue #547, opened 2023-11-01
Clarify replacing hyphens with underscores in id


bbhtt:
The para here states that ***any hyphens in the ID*** should be replaced by underscores to ensure compatibility.

https://github.com/ximion/appstream/blob/d5a1fac9459077a9f96f9a28b272402ccef1c48f/docs/xml/metainfo-component.xml#L98-L101

The ***ID*** is said to be the entirety of `{tld}.{product}.{vendor}` https://github.com/ximion/appstream/blob/d5a1fac9459077a9f96f9a28b272402ccef1c48f/docs/xml/metainfo-component.xml#L84

Is this meant for only the DNS component ie. `{tld}.{vendor}` or the `{product}` component too?

If something is `org.foo.foo-bar` should it be `org.foo.foo_bar`?

The flatpak documentation differs here saying https://docs.flatpak.org/en/latest/conventions.html

> ... as a dash - isn’t allowed except on the last component

Thanks!

bbhtt:
Note that the desktop file spec and DBus spec matches the flatpak documentation which both say to replace it only in the reverse DNS part:

https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#file-naming

> Well-known names containing the dash are allowed but not recommended, because the dash is not allowed in some related uses of reversed DNS names, such as D-Bus object paths and interface names, and Flatpak app IDs. If the author's domain name contains a dash, replacing it with an underscore is recommended: this cannot cause ambiguity, because underscores are not allowed in DNS domain names. 

https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names

> If the author's DNS domain name contains hyphen/minus characters ('-'), which are not allowed in D-Bus interface names, they should be replaced by underscores

ximion:
The AppStream ID should follow the D-Bus spec, so we are a bit too strict here and can relax the "no hyphen" requirement on the last component of the ID.

PR #548, opened 2023-11-02
Relax no-hyphens requirements on the last component of the id tag

bbhtt:
Closes https://github.com/ximion/appstream/issues/547


ximion:
Needs to be adjusted in the validator as well in order to be good to go...

bbhtt:
Ah I thought only the docs change was needed. Not familiar with libappstream code, I'll let you handle it. Sorry!

PR #553, opened 2023-11-08
Explain metainfo-ancient error in more detail

Alexander-Wilms:
I tried to validate https://github.com/LibreSprite/LibreSprite/blob/4dac17fd198288d91c07592373449dcdddaf7e24/desktop/libresprite.appdata.xml and got this error:

```
$ flatpak run org.freedesktop.appstream.cli validate --explain desktop/libresprite.metainfo.xml
E: ~:2: metainfo-ancient
   The metainfo file uses an ancient version of the AppStream specification, which can not be
   validated. Please migrate it to version 0.6 (or higher).

✘ Validation failed: errors: 1
```

I compared it to a valid metainfo file to see the differences at a glance.
I think adding the explanation regarding the different root tag would be helpful.

### DIFF
--- a/src/as-validator-issue-tag.h
+++ b/src/as-validator-issue-tag.h
@@ -772,7 +772,8 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	{ "metainfo-ancient",
 	  AS_ISSUE_SEVERITY_ERROR,
 	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
-	  N_("The metainfo file uses an ancient version of the AppStream specification, which can not be validated. Please migrate it to version 0.6 (or higher), which uses a <component> root tag instead of <application> among other things."),
+	  N_("The metainfo file uses an ancient version of the AppStream specification, which can not be validated. Please migrate it to version 0.6 (or higher). "
+	  "Modern files use the `component` root tag and include many other changes, so check changes carefully when modernizing the data."),
 	},
 
 	{ "root-tag-unknown",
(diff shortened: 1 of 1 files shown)


ximion:
I specifically didn't mention this detail in the message, to prevent people from just renaming the root tag and calling it a day - there's a ton of differences between ancient AppStream and the modern one, so people really need to carefully go through the file and likely fix a bunch of stuff...

src/as-validator-issue-tag.h
```diff
@@ -772,7 +772,7 @@ AsValidatorIssueTag as_validator_issue_tag_list[] =  {
 	{ "metainfo-ancient",
 	  AS_ISSUE_SEVERITY_ERROR,
 	  /* TRANSLATORS: Please do not translate AppStream tag and property names (in backticks). */
-	  N_("The metainfo file uses an ancient version of the AppStream specification, which can not be validated. Please migrate it to version 0.6 (or higher)."),
+	  N_("The metainfo file uses an ancient version of the AppStream specification, which can not be validated. Please migrate it to version 0.6 (or higher), which uses a <component> root tag instead of <application> among other things."),
```
ximion, line 775:
```suggestion
	  N_("The metainfo file uses an ancient version of the AppStream specification, which can not be validated. Please migrate it to version 0.6 (or higher). "
	  "Modern files use the `component` root tag and include many other changes, so check changes carefully when modernizing the data."),
```

Alexander-Wilms:
I was just wondering how it could tell it was an old AppData file, so with the explanation it should be less surprising.

PR #558, opened 2023-11-15
Ensure the stemmer fully loads the init locale once

vivienbiPX:
Hello,

I think there is a bug because the scanner cannot get initialized.

What do you think?

Best regards,

Vivien

### DIFF
--- a/src/as-stemmer.c
+++ b/src/as-stemmer.c
@@ -76,9 +76,9 @@ as_stemmer_init (AsStemmer *stemmer)
 
 	/* we don't use the locale in XML, so it can be POSIX */
 	locale = as_get_current_locale_posix ();
-	stemmer->current_lang = as_utils_locale_to_language (locale);
 
-	as_stemmer_reload (stemmer, stemmer->current_lang);
+	stemmer->current_lang = NULL;
+	as_stemmer_reload (stemmer, locale);
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


ximion:
I do not understand what you are trying to fix here...
Your change will break locale switching and stemming, and cause the stemmer to be reinitialized needlessly as well, so that's probably not what you are trying to achieve. It will also break unit tests that are run on any non-English locale.

vivienbiPX:
I am trying to fix the test failure of the stemmer. It tries to load the "en" language, but since the `as_stemmer_init` function defines `stemmer->current_lang` to `as_utils_locale_to_language (locale)` before calling `as_stemmer_reload`, then the call to `as_stemmer_reload` exits immediately. So the `sb` field is never initialized by the `as_stemmer_init` function.

ximion:
The line `if (as_str_equal0 (lang, stemmer->current_lang))` checks if the locale passed to `as_stemmer_reload` is equal to the stemmer's `current_lang`. If it is, no reload is necessary and we exit. Otherwise we reload the stemmer.

While looking at the code, I did find a different issue though which could lead to the stemmer never being initialized in some edge cases - that's *really* bad and resolved now, maybe that was the thing actually causing your issue?
If not, maybe the POSIX locale doesn't get properly split into its language segment, so you don't initialize the stemmer with a language, maybe?

vivienbiPX:
This is literally the same issue, but if you prefer your solution, I’m fine with that.

ximion:
Ah, sorry! Thank you for finding the issue then and proposing a patch, that is very much appreciated! Your patch removed the `as_utils_locale_to_language` transitions though, which would definitely have broken stemming for many languages. The thing that changed behavior was mostly `stemmer->current_lang = NULL;`, but I rather have the reload happen explicitly than implicitly though a magic string property being NULL.

PR #556, opened 2023-11-12
fix macos build

chenrui333:
closes #551

relates to homebrew/homebrew-core#154003

PR #567, opened 2023-12-10
meson: Pass -D_DARWIN_C_SOURCE on darwin

bobby285271:
This is probably a follow-up for 02ee2d0daa ("fix macos build") and unbreaks the build on Nixpkgs x86_64-darwin. I don't know why, probably due to using even older SDKs? :upside_down_face: 

https://github.com/NixOS/nixpkgs/runs/19486840382
https://logs.ofborg.org/?key=nixos/nixpkgs.273297&attempt_id=05115a28-17b2-4d53-b0e9-d6c2f6a02b63

```
> In file included from ../src/as-system-info.c:60:
> In file included from /nix/store/51c1k4sj0irwgyvkgbyhqi0ggilyh3j2-Libsystem-1238.60.2/include/sys/sysctl.h:83:
> /nix/store/51c1k4sj0irwgyvkgbyhqi0ggilyh3j2-Libsystem-1238.60.2/include/sys/ucred.h:91:2: error: unknown type name 'u_long'
>         u_long  cr_ref;                 /* reference count */
>         ^
> 1 error generated.
> ninja: build stopped: subcommand failed.
```

Update: Instead of adding another typedef, apply the suggestions on pull request 556 to pass `-D_DARWIN_C_SOURCE`, which seems to work fine for us too.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -136,6 +136,11 @@ if host_machine.system() != 'netbsd'
     add_project_arguments('-D_POSIX_C_SOURCE=201710L', language: 'c')
 endif
 
+if host_machine.system() == 'darwin'
+    # See https://github.com/ximion/appstream/issues/551
+    add_project_arguments('-D_DARWIN_C_SOURCE', language: 'c')
+endif
+
 if cc.get_id() == 'clang'
     # Clang doesn't understand autofree helpers on GMutexLocker and thinks
     # these variables are irrelevant, so this warning when used with Clang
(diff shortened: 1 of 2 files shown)


bobby285271:
Oh I just saw https://github.com/ximion/appstream/pull/556#issuecomment-1847176075, give me a few sec to try...

ximion:
Awesome, thanks for the patch! This feels like a much better solution than the previous type redefinitions.

PR #571, opened 2023-12-29
Docs: Mention that <summary/> tag can be translated

Alexander-Wilms:
Not sure if this is correct, but `appstreamcli validate` doesn't complain about it.


ximion:
What an odd oversight! Thanks for the patch and for noticing.
A terse `This element is translatable.` string is probably enough here though, as with most other tags, as how stuff is translated is already defined elsewhere :-)

Alexander-Wilms:
Done

ximion:
Excellent, thank you!

PR #573, opened 2023-12-29
relation-display-length-value-invalid message: Remove mention of shor…

Alexander-Wilms:
…thands


ximion:
Oh, that was an oversight, good catch and thank you for the patch!

PR #591, opened 2024-02-12
qt: Component::customValue should be const

aleixpol:
Since it was omitted, add the const and overload the method for now.

### DIFF
--- a/qt/component.h
+++ b/qt/component.h
@@ -253,7 +253,8 @@ public:
     void setMergeKind(MergeKind kind);
 
     QHash<QString, QString> custom() const;
-    QString customValue(const QString &key);
+    QString customValue(const QString &key) const;
+    [[deprecated]] QString customValue(const QString &key); //TODO Remove when we break ABI
     bool insertCustomValue(const QString &key, const QString &value);
 
     QList<AppStream::ContentRating> contentRatings() const;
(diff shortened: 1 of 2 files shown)


qt/component.h
```diff
@@ -253,7 +253,8 @@ class APPSTREAMQT_EXPORT Component
     void setMergeKind(MergeKind kind);
 
     QHash<QString, QString> custom() const;
-    QString customValue(const QString &key);
+    QString customValue(const QString &key) const;
+    QString customValue(const QString &key); //TODO Remove when we break ABI
```
ximion, line 257: Hmm, can this be marked as `[[deprecated]]`, or will that cause trouble with the C++ compiler not being smart enough?

ximion (changes requested):

qt/component.h
aleixpol: No, added.

We'll just get deprecation warnings in cases where it's used without std::as_const. IIRC c++ prefers the non-const variant.

ximion: > We'll just get deprecation warnings in cases where it's used without std::as_const. IIRC c++ prefers the non-const variant.

That's what I was worried about :-/

The benefit of the deprecation tag is that I will be able to easily spot and remove it once we break API though, a comment is harder to spot (especially when it doesn't contain the keyword "deprecated" ^^)

If there's deprecation warnings that can't be avoided, let me know and we can add a comment instead!

PR #592, opened 2024-02-14
sysinfo: Swap arguments of g_pattern_match_simple

livingsilver94:
According to the [GLib doc](https://docs.gtk.org/glib/func.pattern_match_simple.html) the globbed pattern is the first argument, while the string to test comes after.

This fixes the fact that `as_system_info_has_device_matching_modalias` could return `TRUE` for a 1:1 equality with the pattern, but always returned `FALSE` when using wildcards.


ximion:
That mistake from the Windows port truly is a gift that keeps on giving...
Apparently I didn't find all cases when looking through it. This for sure needs a testcase, but in the meanwhile I see no reason not to merge this.

Thank you for finding this and for fixing it! :-)

PR #601, opened 2024-02-19
Fix doc typos

razzeee:


ximion:
64 is such a beautiful color ;-)
Thanks for the fix!

PR #603, opened 2024-02-20
Use io.github for ID example

TingPing:
Github very specifically puts all user hosted websites under `github.io` as its a completely different domain than `github.com` which controls all officially hosted websites.

Using `io.github` for an ID makes it explicitly clear it is not associated with an official Github project which certainly could be packaged.


TingPing:
No `$name.github.io` can only be created by that User/Organization.

ximion:
I could create a repo with that name, but looks like I can't deploy it.
So in that case, using `io.` is fine :-)

PR #610, opened 2024-02-28
Validate that lists don't get translated

JakobDev:
See https://github.com/flathub/flathub/issues/5005 for reference

### DIFF
--- a/src/as-validator.c
+++ b/src/as-validator.c
@@ -953,11 +953,11 @@ as_validator_check_description_paragraph (AsValidator *validator, xmlNode *node)
 static void
 as_validator_check_description_enumeration (AsValidator *validator, xmlNode *node)
 {
-	g_autofree gchar *lang = NULL;
-	lang = as_xml_get_prop_value (node, "lang");
-	if (lang != NULL) {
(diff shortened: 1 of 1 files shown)


src/as-validator.c
```diff
@@ -953,6 +953,12 @@ as_validator_check_description_paragraph (AsValidator *validator, xmlNode *node)
 static void
 as_validator_check_description_enumeration (AsValidator *validator, xmlNode *node)
 {
+	g_autofree gchar *lang = NULL;
+	lang = as_xml_get_prop_value (node, "lang");
+	if (lang != NULL) {
+		as_validator_add_issue (validator, node, "tag-not-translatable", (const gchar *) node->name);
```
ximion, line 959: Please replace the whole added block with a single
```c
as_validator_check_nolocalized (
	validator,
	node,
	"tag-not-translatable",
	(const gchar *) node->name);
```

ximion (changes requested):
I could have sworn that check existed, but apparently it does only for Catalog metadata...
Anyway, one small change and this should be good to go!

PR #615, opened 2024-03-06
docs: List AppStreamCompose as related to AppStream

tytan652:
Improve discoverability of the AppStreamCompose API in AppStream API docs, in a similar way as it's done between Glib, GIO and GObject in their docs.


ximion:
Very nice, adding this would make a lot of sense rather than just mentioning it on https://www.freedesktop.org/software/appstream/docs/chap-AppStream-API.html :-)

Thanks for the patch!

PR #620, opened 2024-03-16
qt: Fix relocation of Config.cmake files

Johnnynator:
Using the absolute LIBDIR_FULL breaks the use of the cmake files
when cross compiling for a system in a different sysroot.

The ${PACKAGE_PREFIX_DIR} was already used for /include,
reusing it for libs resolves this issue.


ximion:
Makes sense, thank you for the patch!

Issue #611, opened 2024-02-29
Screenshot environment


CarlSchwan:
In https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots

there is an example with `environment="plasma-mobile"`

```
  <screenshot environment="plasma-mobile">
    <caption>The FooBar main window, but on Plasma Mobile</caption>
    <image type="source" width="1600" height="900">https://example.com/foobar/screenshot-1_plasma-mobile.png</image>
  </screenshot>
```

but unfortunately plasma-mobile is not in the authorized styles: https://github.com/ximion/appstream/blob/main/data/desktop-style-ids.txt

ximion:
Yes. So, is this a request to add it to the available styles, or to change the example? ;-)

1Maxnet1:
> Yes. So, is this a request to add it to the available styles, or to change the example? ;-)

I just stumbled across the vary same issue. I would propose to add plasma-mobile as an environment. Screenshots of the mobile appearance of an application might be very different from what they would look on plasma, so in my opinion it makes sense to add it. Let me know, if you can use a little PR for that :)

ximion:
A PR would be welcome (see [desktop-style-ids.txt](https://github.com/ximion/appstream/blob/main/data/desktop-style-ids.txt)). I guess the "mobile" is not a style, so it should indeed be `plasma-mobile` and not `plasma:mobile` (which would also allow `plasma-mobile:dark` later, I guess).

razzeee:
This can be closed

PR #622, opened 2024-03-26
Add Plasma Mobile to desktop-style-ids.txt

1Maxnet1:
As discussed in #611 this PR adds Plasma Mobile to the list of desktop style ids.

Let me know if anything is missing

### DIFF
--- a/data/desktop-style-ids.txt
+++ b/data/desktop-style-ids.txt
@@ -1,20 +1,21 @@
 # List of recognized GUI environments
 # ID            Human-readable name
 #
-cinnamon    Cinnamon
-dde         Deepin
-ede         EDE
-endless     Endless
-gnome       GNOME
-gnome:dark  GNOME (Dark)
-lxde        LXDE
-lxqt        LXQt
-macos       macOS
-mate        Mate
-pantheon    Pantheon
-plasma      KDE Plasma
-razor       Razor
-rox         Rox
-unity       Unity
-windows     Microsoft Windows
-xfce        Xfce
+cinnamon      Cinnamon
+dde           Deepin
+ede           EDE
+endless       Endless
+gnome         GNOME
+gnome:dark    GNOME (Dark)
+lxde          LXDE
+lxqt          LXQt
+macos         macOS
+mate          Mate
+pantheon      Pantheon
+plasma        KDE Plasma
+plasma-mobile KDE Plasma Mobile
+razor         Razor
+rox           Rox
+unity         Unity
+windows       Microsoft Windows
+xfce          Xfce
--- a/data/desktop-style-ids.txt
+++ b/data/desktop-style-ids.txt
@@ -13,7 +13,7 @@ macos         macOS
 mate          Mate
 pantheon      Pantheon
 plasma        KDE Plasma
-plasma-mobile KDE Plasma Mobile
+plasma-mobile Plasma Mobile
 razor         Razor
 rox           Rox
 unity         Unity
--- a/src/as-desktop-env-data.h
+++ b/src/as-desktop-env-data.h
@@ -97,6 +97,8 @@ AsGUIEnvStyleData as_gui_env_style_data[] = {
 	{ "pantheon", N_("Pantheon") },
 	/* TRANSLATORS: Name of the "plasma" visual environment style. */
 	{ "plasma", N_("KDE Plasma") },
+	/* TRANSLATORS: Name of the "plasma-mobile" visual environment style. */
+	{ "plasma-mobile", N_("KDE Plasma Mobile") },
 	/* TRANSLATORS: Name of the "razor" visual environment style. */
 	{ "razor", N_("Razor") },
 	/* TRANSLATORS: Name of the "rox" visual environment style. */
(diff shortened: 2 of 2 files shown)


tintou:
Wouldn't a "chassis" property make more sense? (as defined in https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.hostname1.html#Semantics )
This would be treated as a hint for the screenshot

1Maxnet1:
> Wouldn't a "chassis" property make more sense? (as defined in https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.hostname1.html#Semantics )

I agree introducing a chassis type, makes sense, but would not eliminate the need for plasma-mobile to be added. If you have chassis type desktop/mobile and plasma/plasma-mobile, you could still have Plasma on a desktop/laptop, Plasma on a tablet or convertible (which could prefer the mobile chassis for screenshots, but runs the Desktop Version of plasma), a plasma-mobile installation on a smartphone and plasma-mobile on a docked smartphone (which again would potentially prefer the desktop screenshots as long as you are in a "desktop/docked" mode. To identify all those four cases, one would need both attributes. I admit however that this is currently an edge case. Nevertheless plasma-mobile and plasma are two different desktop environments so ignoring the mentioned edge case it might make sense to add it anywhere.

razzeee:
I guess we should add `phosh` and `gnome-mobile` too then

1Maxnet1:
> I guess we should add `phosh` and `gnome-mobile` too then

I agree for phosh, however gnome-mobile is as far as I know "only" a WIP branch of GNOME which eventually aims to be completely upstreamed. (See the postmarketOS Wiki for details https://wiki.postmarketos.org/wiki/GNOME) so I would not add it to the list. Phosh and Plasma Mobile make sense to me, because they plan to stay separate environments from GNOME/KDE Plasma, while sharing certain components. Would you add Phosh to this MR or should I create another for it?

razzeee:
Still, apps targeting gnome-mobile might have different screenshots (probably portait mode) and it would be useful to know that via metadata. That's why I thought it might be smart to add `gnome-mobile` - that said, you're not wrong.

I think adding `phosh` here would be fine, but @ximion 's call

1Maxnet1:
> Still, apps targeting gnome-mobile might have different screenshots (probably portait mode) and it would be useful to know that via metadata. That's why I thought it might be smart to add `gnome-mobile` - that said, you're not wrong.

Maybe a fitting alternative here would be to add `gnome:mobile` instead as a style. See also the note here: https://github.com/ximion/appstream/issues/611#issuecomment-2020115882

razzeee:
> > Still, apps targeting gnome-mobile might have different screenshots (probably portait mode) and it would be useful to know that via metadata. That's why I thought it might be smart to add `gnome-mobile` - that said, you're not wrong.
> 
> Maybe a fitting alternative here would be to add `gnome:mobile` instead as a style. See also the note here: [#611 (comment)](https://github.com/ximion/appstream/issues/611#issuecomment-2020115882)

I think that would be bad as it probably breaks the possibility to have `gnome-mobile:dark` then

data/desktop-style-ids.txt
```diff
@@ -1,20 +1,21 @@
...
+endless       Endless
+gnome         GNOME
+gnome:dark    GNOME (Dark)
+lxde          LXDE
+lxqt          LXQt
+macos         macOS
+mate          Mate
+pantheon      Pantheon
+plasma        KDE Plasma
+plasma-mobile KDE Plasma Mobile
```
CarlSchwan, line 16:
```suggestion
plasma-mobile Plasma Mobile
```

While for Plasma desktop we often add "KDE" in front, for Plasma Mobile we don't do that anywhere.

src/as-desktop-env-data.h
```diff
@@ -97,6 +97,8 @@ AsGUIEnvStyleData as_gui_env_style_data[] = {
 	{ "pantheon", N_("Pantheon") },
 	/* TRANSLATORS: Name of the "plasma" visual environment style. */
 	{ "plasma", N_("KDE Plasma") },
+	/* TRANSLATORS: Name of the "plasma-mobile" visual environment style. */
+	{ "plasma-mobile", N_("KDE Plasma Mobile") },
```
CarlSchwan, line 101:
```suggestion
	{ "plasma-mobile", N_("Plasma Mobile") },
```

CarlSchwan (commented):
Also I agree with the others we should at least add phosh and probably also gnome mobile too, to that list :)

ximion:
> Also I agree with the others we should at least add phosh and probably also gnome mobile too, to that list :)

I do no longer just add stuff because it might be used in future, but only when it is specifically requested by the respective project owners / maintainers / contributors.
There were some instances in the past where we preemptively added stuff to AppStream that the respective projects had different ideas about, that's why I am fairly conservative (we can always add stuff, but once IDs are in, removing them will be very hard or even impossible).

ximion:
Also, that C header is auto-generated, it doesn't need to be changed specifically as it will be regenerated at the next opportunity (but changing it also doesn't hurt in this case, so with respect to that, this patch is fine).

razzeee:
Sorry for the noise, please don't add phosh and gnome mobile then

1Maxnet1:
Thanks for all the comments. I updated the PR with @CarlSchwan's suggestions and as mentioned by @ximion I won't add Phosh or GNOME Mobile for now.

ximion:
Nice! Let's merge it then! :-)

PR #632, opened 2024-05-16
Warn in the documentation that support for the legacy path has been dropped

rautyrauty:
/usr/share/app-info/(xml|xmls) are no longer scanned

### DIFF
--- a/docs/xml/catalog-xmldata.xml
+++ b/docs/xml/catalog-xmldata.xml
@@ -37,9 +37,8 @@
 		<important>
 			<title>Legacy Path</title>
 			<para>
-				AppStream tools scan the paths <filename>/usr/share/app-info/(xml|xmls)</filename> / <filename>/var/lib/app-info/(xml|xmls)</filename> path for legacy
-				compatibility as well. If possible, the old locations and old layouts should not be used anymore.
-				Support for the legacy path will likely be dropped completely with a future AppStream 1.0 release.
+				Prior to version 1.0, AppStream tools scanned the paths <filename>/usr/share/app-info/(xml|xmls)</filename> and <filename>/var/lib/app-info/(xml|xmls)</filename> path for legacy
+				compatibility as well.  Legacy path support was dropped in version 1.0. The old locations should not be used anymore. The modern locations are supported by both the AppStream 1.x as well as AppStream 0.16.x series.
 			</para>
 		</important>
 	</section>


ximion:
Hmm, actually support shouldn't be dropped yet - there should be a symlink in place from `app-info` -> `swcatalog`. But sure, adding a warning is a good idea, because support will likely be gone completely in a few releases.

rautyrauty:
> Hmm, actually support shouldn't be dropped yet - there should be a symlink in place from `app-info` -> `swcatalog`. 

Creating a symlink is not a support for scanning these directories.

> But sure, adding a warning is a good idea, because support will likely be gone completely in a few releases.

Thanks.

docs/xml/catalog-xmldata.xml
```diff
@@ -34,14 +34,13 @@
...
 		</para>
 
-		<important>
+		<warning>
 			<title>Legacy Path</title>
 			<para>
-				AppStream tools scan the paths <filename>/usr/share/app-info/(xml|xmls)</filename> / <filename>/var/lib/app-info/(xml|xmls)</filename> path for legacy
-				compatibility as well. If possible, the old locations and old layouts should not be used anymore.
-				Support for the legacy path will likely be dropped completely with a future AppStream 1.0 release.
+				Prior to version 1.0.3, AppStream tools scan the paths <filename>/usr/share/app-info/(xml|xmls)</filename> / <filename>/var/lib/app-info/(xml|xmls)</filename> path for legacy
```
ximion, line 40:
```suggestion
				Prior to version 1.0, AppStream tools scanned the paths <filename>/usr/share/app-info/(xml|xmls)</filename> and <filename>/var/lib/app-info/(xml|xmls)</filename> path for legacy
```

docs/xml/catalog-xmldata.xml
```diff
@@ -34,14 +34,13 @@
...
+				compatibility as well.  Legacy path support was dropped in version 1.0.3. The old locations and old layouts should not be used anymore.
```
ximion, line 41:
```suggestion
				compatibility as well.  Legacy path support was dropped in version 1.0. The old locations should not be used anymore. The modern locations are supported by both the AppStream 1.x as well as AppStream 0.16.x series.
```

docs/xml/catalog-xmldata.xml
```diff
@@ -34,14 +34,13 @@
 			quite large.
 		</para>
 
-		<important>
+		<warning>
```
ximion, line 37: Deprecation into is usually an `important` message box in the spec, not a `warning`. Important messages are also orange, so really hard to miss, while `warning` boxes are rendered bright red and usually in the spec are used to highlight some common pitfall or trait in the active, non-legacy spec that users must be aware of or pay special attention to (while "don't use legacy things" isn't one of those).

rautyrauty: Yes, you're right.

ximion (approved):

ximion:
Excellent, thank you!

PR #633, opened 2024-05-19
i18n: update Catalan translation

juxuanu:
Translated from https://l10n.gnome.org/vertimus/appstream/main/po/ca/. I see some original strings are different from the last version of the ca.po file. Let me know if there is something else needed to be done.


juxuanu:
This needs to go through Weblate. I just did that https://hosted.weblate.org/projects/appstream/translations/ca/.

Closing.

PR #641, opened 2024-07-09
validator: Make sure we read an initialised variable

aleixpol:
We were not reading the kind so it was left undeclared.

src/as-validator.c:3534:7: warning: variable 'ckind' is uninitialized when used here [-Wuninitialized]


ximion:
We do compile with `-Werror` on the CI... It's crazy that this has not been caught for years, apparently!
Which GCC/Clang version are you using?

PR #650, opened 2024-08-15
validator: Omit date validation for snapshot releases

### LINKED ISSUE #570: How can app handle unrelased release date in metada
I've noticed that some app developers use a metadata file to prepare release notes for versions that have not yet been released.

```
      <release version="3.1.0" date="2023-12-XX">
        <description translatable="no">
          <p>This is a minor release of Impression with a few small improvements.</p>
        </description>
      </release>
```

Of course, appstreamcli can't validate these dates.

```
W: @app-id@:68: invalid-iso8601-date 2023-12-XX
   The AppStream specification requires a complete, ISO 8601 date string with at least
   day-granularity to denote dates. Please ensure the date string is valid.
```

Could you offer me a way that developers can use a metadata file to prepare release notes without breaking validation?

jayaddison:
Resolves #570.


ximion:
This is not quite what I meant in the bug report...
*If* there is a `date` defined, then we should absolutely validate it. But for snapshot releases, having no date shouldn't be an error.

Since there may be a release tomorrow I may just implement that change quickly (but not if the release is deferred to Thursday).

jayaddison:
That.. would make a lot more sense, thank you for catching that.

jayaddison:
After re-reading your comment: I'm not sure the adjusted logic I've pushed is correct, either; it could still make sense to provide a non-error-severity output, instead of nothing at all.

jayaddison:
Question: is there a way to call `as_validator_add_issue` while overriding the default severity for a rule?  (idea: `release-time-missing` is usually `error`-level, but we could downgrade it to `warning` for snapshot releases)

Answer (afaict): not currently.

Adding that feature would probably be more-involved than I'm capable of at the moment (C/C++ isn't really my main skillset).  But I could attempt it at some point (perhaps as a separate pull request, to precede merge of this one) if that'd be worthwhile.

jayaddison:
Hm.  There's a possibly-unicode-related error appearing in these builds; I did encounter that locally, but suspected that it might be a difference in my local build environment.  It appears not.

I'm also unsure why the Fedora Rawhide continuous integration build appears as successful here; it should have failed, since the same problem occurred there according to the logs.

jayaddison:
> Hm. There's a possibly-unicode-related error appearing in these builds; I did encounter that locally, but suspected that it might be a difference in my local build environment. It appears not.

Possibly due to repeated (duplicate) evaluation of the `G_GNUC_PRINTF` macro?  Perhaps it makes more sense only to add that macro annotation to the inner helper function.

jayaddison:
> I'm also unsure why the Fedora Rawhide continuous integration build appears as successful here; it should have failed, since the same problem occurred there according to the logs.

`continue-on-error` is set to `true` for those workflows jobs.

jayaddison:
> > Hm. There's a possibly-unicode-related error appearing in these builds; I did encounter that locally, but suspected that it might be a difference in my local build environment. It appears not.
> 
> Possibly due to repeated (duplicate) evaluation of the `G_GNUC_PRINTF` macro? Perhaps it makes more sense only to add that macro annotation to the inner helper function.

I _think_ it's fine to leave those annotations on the outer functions too -- as [documented in GLib](https://docs.gtk.org/glib/func.GNUC_PRINTF.html), it's a macro to assist the compiler with type-checking.

I've rebased the changes against master to reduce the chance that the `insecure-url` warning is confused with some mainline changes related to HTTPS (these are, as far as I can tell, unrelated).

jayaddison:
> My next guess would be that something about the argument-passing related to `va_list` could be problematic. I might try refactoring to remove that.

Yes, this appears to have resolved the problem.  It means there is more repeated/duplicate code than I would like, but it is a start.

jayaddison:
@ximion I think that this changeset now adheres more-closely to what you suggested in #570: specifically that although `release-time-missing` should not be reported as an _error_ for `snapshot` releases, we should still run the relevant validation and inform users of possible problems.

jayaddison:
@ximion my communication on this thread was somewhat verbose, so I'll try to be concise: I'd be glad for any code review on the approach currently coded here (approximately: refactor the issue-emission code to allow a severity to be passed by-argument, and use that functionality to emit a reduced-severity (warning instead of error) notice for `release-time-missing` in the specific case of `snapshot` releases).

jayaddison:
Sorry to nag: is this likely to be reviewable in the nearish future?  I'm very much not a competent C++ developer, so if the changes initially seem rough/unacceptable, I'd probably prefer to close it and await a more robust implementation.

jayaddison:
Closing to reduce my mental overhead of open tasks -- not discouraged about reaching a solution for this, just trying to refocus.  My eventual hope with resolution of #570 is that I can re-enable the `release-time` validation for `gnome-maps` ([disabled](https://gitlab.gnome.org/GNOME/gnome-maps/-/commit/27bb4c258a96a85a747e671f98b290c5d6c199e6) (hopefully temporarily!) as the warnings conflicted with a [build-reproducibility](https://gitlab.gnome.org/GNOME/gnome-maps/-/issues/729) fix).

ximion:
I am very sorry for getting back to this so late and not supervising this PR properly. I had to take some absence from FOSS development for personal reasons.

But the actual change was trivial, sorry that you put so much more work into it than necessary! It will ship with the next release :-)

jayaddison:
No problem at all - real life happens, and should take priority.

I'm very grateful for 13a94d586f2dfed4de61e5986f3ef423837f11d6 - yep, I got a bit frantic attempting an implementation, but it was fun to try some C again, and I learned some things :)  Thanks again!

PR #653, opened 2024-08-21
trivial: Reword guides to guide

sidt4:
Reword guides to guide.

### DIFF
--- a/compose/asc-hint-tags.c
+++ b/compose/asc-hint-tags.c
@@ -59,7 +59,7 @@ AscHintTagStatic asc_hint_tag_list[] =  {
 	{ "ancient-metadata",
 	  AS_ISSUE_SEVERITY_WARNING,
 	  "The AppStream metadata should be updated to follow a more recent version of the specification.<br/>"
-	  "Please consult <a href=\"https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html\">MetaInfo quickstart guide</a> for "
+	  "Please consult the <a href=\"https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html\">MetaInfo quickstart guides</a> for "
 	  "more information."
 	},
 


ximion:
This is incorrect though, as these links point to multiple guides, specifically to the landing page that has guides for addon maintainers, gui app maintainers, distributors, translators, ...

So, I don't think this is a correct change.

sidt4:
It's not clear as it's the same link in https://github.com/ximion/appstream/commit/060991c22f739cd416726c3a0e72357556e796d4 ?

ximion:
> It's not clear as it's the same link in [060991c](https://github.com/ximion/appstream/commit/060991c22f739cd416726c3a0e72357556e796d4) ?

Yeah, I guess my comment to change that was actually wrong :sweat_smile: 
It should have been "guides" there as well, I apparently misread that comment.

sidt4:
Okay. Updated.

compose/asc-hint-tags.c
```diff
@@ -59,7 +59,7 @@ AscHintTagStatic asc_hint_tag_list[] =  {
 	{ "ancient-metadata",
 	  AS_ISSUE_SEVERITY_WARNING,
 	  "The AppStream metadata should be updated to follow a more recent version of the specification.<br/>"
-	  "Please consult <a href=\"https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html\">MetaInfo quickstart guide</a> for "
+	  "Please consult <a href=\"https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html\">MetaInfo quickstart guides</a> for "
```
ximion, line 62:
```suggestion
	  "Please consult the <a href=\"https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html\">MetaInfo quickstart guides</a> for "
```

README.md
```diff
@@ -34,7 +34,7 @@ You can [download signed release tarballs of AppStream on freedesktop.org](https
 ## Information For Users
 
 ### For Upstream Application Authors
-- **[Quickstart Guides](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html)**:
+- **[MetaInfo Quickstart Guides](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html)**:
```
ximion, line 37: The guides have nothing to do with just MetaInfo data - one is explicitly about how to deploy AppStream data if you are a distributor, the other is about translations, and only two are about MetaInfo files. So I think the previous sentence in that context is good enough.

ximion (changes requested):

README.md
sidt4: But the heading in the page says `Metadata Quickstart`. Should that be updated too ?

appstream

ximion: I don't think so... All of this is concerning metadata...

ximion (approved):

PR #655, opened 2024-08-22
ci: Update to Ubuntu 24.04 for Debian Testing builds

### LINKED ISSUE #654: Bug: Debian Testing build fails due to fchmodat2 syscall during documentation build
The `Debian Testing` continuous integration workflow includes a version of `glibc` that uses the `fchmodat2` syscall to provide some extended support for symlink permissioning.

The `fchmodat2` syscall isn't available in the `podman` default `seccomp.json` policy, nor is it supported by `libseccomp2` version 2.5.3 that is currently installed during these continuous integration workflows on Ubuntu 22.04 GitHub Actions runner images.

This means that the syscall returns an error when `daps`, used to build HTML documentation, [invokes `tar`](https://github.com/openSUSE/daps/blob/30e43ef03fa145fc151cc124290c6fa733ee8252/make/html.mk#L221-L223) in a way that invokes `fchmodat2` -- and that we can't simply add it to a custom `seccomp.json` file, because without sufficient `libseccomp2` support (2.5.5 or later), the syscall is unavailable for use in policy files.

Commit bf6a2a9299cef80c0f19ac552dbd396366eb1373 has worked around this temporarily by running the relevant `podman` steps in an `unconfined` seccomp environment -- but I think we should revert that when an updated `libseccomp2` version becomes available in the runners.

See also https://bugs.launchpad.net/ubuntu/+source/tar/+bug/2059734/ for some detailed discussion of the problem in Ubuntu's upstream bugtracker.

Edit: refer to specific Ubuntu release version number.
Edit: add context about `daps`, including a source code reference.

jayaddison:
Resolves #654.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -23,7 +23,7 @@ jobs:
 
   build-debian-testing:
     name: Debian Testing
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v3
 
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -31,8 +31,7 @@ jobs:
       run: cd tests/ci/ && podman build -t appstream -f ./Dockerfile-debian-testing .
 
     - name: Build
-      run: podman run -a stdout -a stderr --security-opt seccomp=unconfined
-           -e COMPILER_VENDOR=$CVENDOR -e CC=gcc -e CXX=g++ -v `pwd`:/build appstream
+      run: podman run -a stdout -a stderr -e COMPILER_VENDOR=$CVENDOR -e CC=gcc -e CXX=g++ -v `pwd`:/build appstream
            ./tests/ci/run-build.sh
 
     - name: Test


jayaddison:
A [patch](https://bugs.launchpad.net/ubuntu/+source/tar/+bug/2059734/comments/16) has been posted to resolve this problem in `jammy` via the backport process, so this distribution-level upgrade may not be required.

jayaddison:
> A [patch](https://bugs.launchpad.net/ubuntu/+source/tar/+bug/2059734/comments/16) has been posted to resolve this problem in `jammy` via the backport process, so this distribution-level upgrade may not be required.

Note / clarification: that patch hasn't reached `jammy` yet, but if+when it does, then the `seccomp` `unconfined` change could be reverted, and this pull request could be closed.

jayaddison:
The patch hasn't reached the latest 22.04.5 release of `jammy`, so on balance, I think I'd prefer to move this back into ready-for-review status (which if I remember correctly, implies mergable).  I'm not sure whether it's better to merge, or wait for a subsequent `jammy` update, or to continue on as-is with the existing `unconfined` behaviour -- but doing this should provide the most options.

ximion:
Nice, thanks!

jayaddison:
@ximion not urgent, but a note that I think it'll be possible (and consistent) to revert this in the near future; specifically when `ubuntu-latest` begins to refer to Ubuntu 24.04, as it should soon ( https://github.com/actions/runner-images/issues/10636 ).  I'll keep a mental note to create a revert PR after that migration is complete (anticipated 2025-01-17).

PR #658, opened 2024-08-31
sysinfo: Add memory size detection support for GNU/Hurd

pinotree:


pinotree:
@sthibaul while this seems to work, would you please take a look in case I'm doing anything wrong? thanks!

ximion:
It looks sensible to me (but I know nothing about Hurd), so, is this safe to merge in your opinion?

sthibaul:
Yes, it's good to merge

ximion:
Nice, thank you! :-)

Issue #646, opened 2024-08-07
AppStream documentation describes "desktop applications" in too restrictive of a manner.


RokeJulianLockhart:
As previously reported at [`gitlab.freedesktop.org/freedesktop/freedesktop/-/issues/1272`](https://gitlab.freedesktop.org/freedesktop/freedesktop/-/issues/1272#:~:text=AppStream%20documentation%20describes%20%22desktop%20applications%22%20in%20too%20restrictive%20of%20a%20manner.):

<blockQuote cite="https://gitlab.freedesktop.org/freedesktop/freedesktop/-/issues/1272#:~:text=AppStream%20documentation%20describes%20%22desktop%20applications%22%20in%20too%20restrictive%20of%20a%20manner.">

##### **Preamble**

[`gitlab.freedesktop.org/freedesktop/freedesktop`](https://gitlab.freedesktop.org/freedesktop/freedesktop#:~:text=issues%20with%20and%20documentation%20for%20our%20infrastructure%2C%20organisation%2C%20services%2C%20and%20processes) states:

> Issues with and documentation for our infrastructure, organisation, services, and processes.

...so I expect that here is the place to discuss documentation.

##### **Issue**

[`freedesktop.org/software/appstream/docs/sect-Metadata-Application.html#spec-appdata-introduction`](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html#spec-appdata-introduction:~:text=a%20desktop%20application%20is%20an%20application%20which%20has%20a%20graphical%20user%20interface%20and%20is%20commonly%20used%20with%20mouse%20and%20keyboard.%20it%20also%20ships%20a%20freedesktop%20.desktop%20file%20to%20be%20visible%20in%20application%20menus.%20) states:

> A desktop application is an application which has a graphical user interface and is commonly used with mouse and keyboard. It also ships a Freedesktop .desktop file to be visible in application menus.

This would exclude all primarily capacitive input-based GUIs, of which *many* exist - anything designed primarily for smartphone usage, especially. Being more nondescript about the input method by stating something like "and is commonly used with mouse, keyboard or touch device" might be better.

</blockQuote>

RokeJulianLockhart:
> https://github.com/ximion/appstream/issues/646#issuecomment-2297464072

@ximion, I'll be glad to. Do you know where the documentation is stored, in source form?

RokeJulianLockhart:
> https://github.com/ximion/appstream/issues/646#issuecomment-2323395942

[...]

PR #659, opened 2024-09-01
Update `metainfo-desktopapp.xml` to better explain desktop applications.

RokeJulianLockhart:
For [this thread](https://github.com/ximion/appstream/issues/646#issuecomment-2297464072:~:text=This%20is%20fairly%20old%20text%20from%202012%20%2D%20patches%20are%20welcome%20to%20update%20it%20:%2D\)):

> This is fairly old text from 2012 - patches are welcome to update it

### DIFF
--- a/docs/xml/metainfo-desktopapp.xml
+++ b/docs/xml/metainfo-desktopapp.xml
@@ -10,9 +10,8 @@
 	<section id="spec-appdata-introduction">
 		<title>Introduction</title>
 		<para>
-			A desktop application is a software package which includes a graphical user interface as its primary method of interaction and is designed to be used with mouse and keyboard,
-			irrespective of whether it also supports different invocation target sizes for alternative methods of input, like capacitive touch devices. Its package must include a Freedesktop
-			<filename>.desktop</filename> file to be visible in application menus.
+			A desktop application is interactive software that presents a graphical interface to the user.
+			To appear in application listings, the packaged application should include a <filename>.desktop</filename> file.
 		</para>
 		<para>
 			AppStream generators may pull data from the preexisting <filename>.desktop</filename> files to represent an application in the AppStream metadata pool. Upstream projects should
--- a/docs/xml/metainfo-desktopapp.xml
+++ b/docs/xml/metainfo-desktopapp.xml
@@ -10,7 +10,8 @@
 	<section id="spec-appdata-introduction">
 		<title>Introduction</title>
 		<para>
-			A desktop application is an application which has a graphical user interface and is commonly used with mouse and keyboard. It also ships a Freedesktop
+			A desktop application is a software package which includes a graphical user interface as its primary method of interaction and is designed to be used with mouse and keyboard,
+			irrespective of whether it also supports different invocation target sizes for alternative methods of input, like capacitive touch devices. Its package must include a Freedesktop
 			<filename>.desktop</filename> file to be visible in application menus.
 		</para>
 		<para>
--- a/docs/xml/metainfo-desktopapp.xml
+++ b/docs/xml/metainfo-desktopapp.xml
@@ -11,7 +11,7 @@
 		<title>Introduction</title>
 		<para>
 			A desktop application is interactive software that presents a graphical interface to the user.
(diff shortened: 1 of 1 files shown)


docs/xml/metainfo-desktopapp.xml
```diff
@@ -10,7 +10,8 @@
 	<section id="spec-appdata-introduction">
 		<title>Introduction</title>
 		<para>
-			A desktop application is an application which has a graphical user interface and is commonly used with mouse and keyboard. It also ships a Freedesktop
+			A desktop application is a software package which includes a graphical user interface as its primary method of interaction and is designed to be used with mouse and keyboard,
+			irrespective of whether it also supports different invocation target sizes for alternative methods of input, like capacitive touch devices. Its package must include a Freedesktop
 			<filename>.desktop</filename> file to be visible in application menus.
```
jayaddison, line 15:
```suggestion
			A desktop application is interactive software that presents a graphical interface to the user.
			To appear in application listings, the packaged application should include a <filename>.desktop</filename> file.
```

Related: [AppStream 1.0.3](https://lists.freedesktop.org/archives/appstream/2024-April/000417.html) added support for `plasma-mobile` as a desktop style -- so the term `desktop` as used here does indeed cater to more devices than those with mouse+keyboard.

docs/xml/metainfo-desktopapp.xml
```diff
@@ -10,8 +10,8 @@
 	<section id="spec-appdata-introduction">
 		<title>Introduction</title>
 		<para>
-			A desktop application is an application which has a graphical user interface and is commonly used with mouse and keyboard. It also ships a Freedesktop
-			<filename>.desktop</filename> file to be visible in application menus.
+			A desktop application is interactive software that presents a graphical interface to the user.
+			To appear in application listings, the packaged application should include a <filename>.desktop</filename> file.
```
jayaddison, line 14: I think I understand now that while the Freedesktop `.desktop` file is essential for processing/inclusion by AppStream, the metainfo files described in the following paragraph are optional extra information about distributed software.

As a result I wonder about two details here:

  * Perhaps we should indeed use the suggested word `must` rather than `should` (it's stricter terminology -- but for an audience already interested in `appstream`, that's probably what we want?).
  * It could be worth restoring the word `Freedesktop` to help people discover the [relevant format documentation](https://specifications.freedesktop.org/desktop-entry-spec/).  In fact, if the XSLT works OK with the nested XML content, let's try that as a hyperlink.

```suggestion
			To appear in application listings, the packaged application must include a <ulink url="https://specifications.freedesktop.org/desktop-entry-spec/latest/">Freedesktop <filename>.desktop</filename></ulink> file.
```

(this also makes me notice that some links elsewhere in the documentation reference a legacy/now-redirected `standards.freedesktop.org` domain)

RokeJulianLockhart: > (this also makes me notice that some links elsewhere in the documentation reference a legacy/now-redirected standards.freedesktop.org domain)

@jayaddison, if it would be of use, I'd accept changes across disparate files to fix that problem in this PR, but I imagine that the repository maintainers would prefer us to try that in a separate PR.

Would that be as simple as a F&R? It's indeed generally best to reduce the amount of 301s.

jayaddison: > > (this also makes me notice that some links elsewhere in the documentation reference a legacy/now-redirected standards.freedesktop.org domain)
> 
> @jayaddison, if it would be of use, I'd accept changes across disparate files to fix that problem in this PR, but I imagine that the repository maintainers would prefer us to try that in a separate PR.

Agreed, if I were in the shoes of a maintainer, that would be my preference.
 
> Would that be as simple as a F&R? It's indeed generally best to reduce the amount of `301`s.

_Possibly_.  We should confirm that the URL path structure does not change -- or alternatively check for broken links before and after such a migration.

jayaddison: URL-update suggestions opened as #663.

ximion:
Sorry for the delay!

PR #660, opened 2024-09-03
MetaInfo: typo fixup

jayaddison:
Noticed while reading more of the documentation to improve my understanding of the context for #659.

cc @RokeJulianLockhart (thanks for accepting my changes in that PR)


jayaddison:
I think I'd like to offer to rewrite all four of the paragraphs in this XML file; until I'm confident about that, though, I'd prefer to begin with fixups/typo corrections.

RokeJulianLockhart (approved):

PR #666, opened 2024-09-18
sysinfo: Add memory size detection support for illumos and Solaris

mtelka:


ximion:
I can't properly validate this as I am not knowledgeable in Solaris, but the patch looks sound.
Thank you for the addition!

PR #669, opened 2024-09-28
Add COSMIC to xdg-category-names.txt

jackpot51:
COSMIC is a desktop environment and a valid category name as part of the specification: https://specifications.freedesktop.org/menu-spec/latest/additional-category-registry.html

### DIFF
--- a/data/xdg-category-names.txt
+++ b/data/xdg-category-names.txt
@@ -32,6 +32,7 @@ ConsoleOnly
 Construction
 ContactManagement
 Core
+COSMIC
 DDE
 DataVisualization
 Database


jackpot51:
@ximion I would strongly request that COSMIC is not filtered our of the metadata as "low quality", we are going to use it to show applications and applets specifically made for the platform.

razzeee:
I would recommend using keywords

razzeee:
Related https://github.com/flathub-infra/website/issues/3201

ximion:
> @ximion I would strongly request that the COSMIC category is not filtered out of the metadata as "low quality", we are going to use it to show applications and applets specifically made for the platform.

We will phase out *all* desktop and toolkit specific categories in the near future, and I am quite against adding even more. If you want to filter things by a project/organization, you can use https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-project_group

jackpot51:
Ok, I will recommend using project_group. My concern with that is that it only allows one value.

jackpot51:
> We will phase out all desktop and toolkit specific categories in the near future, and I am quite against adding even more

Where is this being discussed? COSMIC is part of the spec now so it does not seem valid to refuse to include it.

razzeee:
All other DEs are also filtered afaik

jackpot51:
We have come up with an alternative.

PR #670, opened 2024-09-29
qt: Expose branding color

CarlSchwan:


aleixpol:
Note this breaks the FreeBSD CI because the forward declaration is by class rather than struct.

I'd pass this values as QStringView rather than QString when possible (not that it matters much, most of these APIs will be unused).

qt/branding.h
```diff
@@ -0,0 +1,88 @@
+// SPDX-FileCopyrightText: 2024 Carl Schwan <carl@carlschwan.eu>
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+#pragma once
+
+#include <QObject>
+#include <QSharedDataPointer>
+#include "appstreamqt_export.h"
+
+class _AsBranding;
```
ximion, line 10: This should be `struct`, shouldn't it?

qt/branding.cpp
```diff
@@ -0,0 +1,102 @@
+// SPDX-FileCopyrightText: 2024 Carl Schwan <carl@carlschwan.eu>
+// SPDX-License-Identifier: LGPL-2.1-or-later
```
ximion, line 2: Can you use the full copyright please, like in all other files? I do like the SPDX strings, but I think it will be easier to convert everything all at once at a future point, rather than having one outlier file.

qt/branding.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    AsBranding *brandingp;
+};
+
+Branding::Branding()
+    : d(new BrandingData)
+{
+}
+
+Branding::Branding(_AsBranding *devp)
```
ximion, line 42: `devp` should probably be `branding`

ximion (changes requested):

CarlSchwan:
> Note this breaks the FreeBSD CI because the forward declaration is by class rather than struct.
> 
> I'd pass this values as QStringView rather than QString when possible (not that it matters much, most of these APIs will be unused).

QStringView doesn't have a qPrintable overload, so not really worth it.

ximion (approved):

ximion:
Thanks! :-)

PR #671, opened 2024-09-30
qt: Specify utc as the dates' timezone

aleixpol:
Otherwise it loads all the dates as if they were in the local timezone, making it all a bit more complex for no reason.


ximion:
Since the timestamp is UNIX time, I think this is okay (ISO8601 time may contain a timezone, but I guess if we convert that into a timestamp, we parse that correctly already).

So, this looks good to go, thanks!

PR #675, opened 2024-10-30
bundle: Add AS_BUNDLE_KIND_SYSUPDATE

Kekun:
This is needed for GNOME Software to have a correct bundle type for systemd-sysupdate.

### DIFF
--- a/docs/xml/catalog-xmldata.xml
+++ b/docs/xml/catalog-xmldata.xml
@@ -598,6 +598,7 @@
 							<listitem><para><code>tarball</code> - For plain and possibly compressed tarballs.</para></listitem>
 							<listitem><para><code>cabinet</code> - For cabinet firmware deployments.</para></listitem>
 							<listitem><para><code>linglong</code> - For <ulink url="https://linglong.dev/">Linglong</ulink> bundles.</para></listitem>
+							<listitem><para><code>sysupdate</code> - For <ulink url="https://www.freedesktop.org/software/systemd/man/latest/systemd-sysupdate.html">systemd-sysupdate</ulink> bundles.</para></listitem>
 						</itemizedlist>
 					</para>
 					<para>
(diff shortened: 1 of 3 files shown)


ximion:
Man, we are really stretching the definition of "bundle" now - software delivery method is more apt ^^

But I have nothing against this in principle, given what we use `bundle` for already, sysupdate fits in neatly.
Thanks for the patch!

AdrianVovk:
Maybe instead of having a single bundle for sysupdate, which you correctly describe as a delivery mechanism more than a bundle format, we could have a couple different bundle types that may fit a bit better. sysupdate manages the following kinds of bundles:

- os images. This is akin to the `package` bundle type, but it's for a whole OS distributed as a single image (including ostree, sysupdate, bootc, Vanilla's updater, etc)
- portable services. These are bundles that contain a system service
- sysexts. These are bundles that get merged into the OS. Maybe they belong as part of an os image bundle? IDK
- configuration extensions. These are bundles of settings, provisioned by a system administrator.
- VMs and containers. These are again OS images, but they're not for the host system at all.
- "components", which can be basically anything in any format. Components are distributed as part of the OS, however

I'd boil this down into:
- OS_IMAGE, for the OS image itself (kernel, /usr tree, etc). As opposed to `package`. Also containers and VMs are here too, since that's all sysupdated can manage for them
- OS_COMPONENT, for somewhat independent parts of the OS image. sysexts, sysupdate components, etc. Again, as opposed to `package`
- PORTABLE_SERVICE: portable services
- CONFEXT: confexts

Marking this all as just a "sysupdate" bundle isn't a bad solution either, though :shrug:. Seems like you already cut a release with this patch, so it's probably not worth touching anymore. But, if you do feel like touching it you could probably just map the sysupdate bundle kind to the os_image bundle kind for backwards compat

PR #685, opened 2024-11-19
as-vercmp: Fix gtkdoc to not start with '>'

sidt4:
This causes rendering issues in web browser.

Refer: https://www.freedesktop.org/software/appstream/docs/api/func.vercmp_simple.html

### Before:

before

### With Fix:

after


ximion:
Fun...
Thanks for the patch!

PR #687, opened 2024-11-30
Fix some tests on Windows

QuLogic:
As of https://github.com/ximion/appstream/pull/510, fetching the Kernel version is not supported on Windows.


ximion:
This is fine for now, as it fixes the tests that will never pass as the respective functionality is not implemented on Windows.
In the long run though, it would actually be nice to get Windows properly supported...

Thanks for the patch!

QuLogic:
Unfortunately, I couldn't find a consistent API for doing so. They have `GetVersionEx`, but it's deprecated, and follows the application manifest, not Windows (so if you manifest for Windows 8, you get 8, regardless of where it's running.) And then a bunch of `IsVersionXYZOrGreater` functions...

PR #691, opened 2025-01-13
Add android and iOS

razzeee:
@CarlSchwan mentioned, that this would be helpful for KDE apps


ximion:
I think it makes sense to add this - we have Plasma Mobile and Phosh as well, after all.

PR #692, opened 2025-01-14
meson: Allow tests to be optional

bbhtt:


meson_options.txt
```diff
@@ -65,6 +65,16 @@ option('install-docs',
        value : true,
        description : 'Install documentation for API and specification'
 )
+option('tests',
+       description: 'Build the test suite',
+       type: 'boolean',
+       value: true
+)
```
eli-schwartz, line 72: Tests are always optional, you never *have* to run them. What's the goal here? Just to avoid compiling the test programs by default?

I would suggest instead setting `build_by_default: false` for test progs. Meson 1.7 won't build them by default anymore, if you do so.

bbhtt: Yes 

1.7 is too new for me to benefit from it.

ximion:
I think I will have to veto this. I *always* want the tests to be compiled - to at least have them build is the easiest way to find any issues early instead of shipping something broken onto systems which might then break in subtle ways.

Of course, people may not run the tests after build, but that risk is up to the people building the code ;-)

Is there any reason why you don't want to even build the tests? They are extremely quick to compile and quite lightweight.

eli-schwartz:
TBH I'm not sure I understand the distinction. Distributors want to run tests after build too, obviously, as they don't like regressions either. But it's possible to toggle them off if e.g. you've already run it (or someone else has), and skipping them saves you on 24 compile jobs (8 tests times 2 sources + 1 link stage).

It may not be the biggest deal in the world, but it's not nothing.

Again, tagging them as `build_by_default: false` means they get compiled just-in-time for the test phase itself, no need to bother people with additional configure options and the tests are always enabled for anyone that tries running it. It's not clear to me that simply compiling the test programs provides valuable additional testsuite coverage...

bbhtt:
> I always want the tests to be compiled

That is what it is doing by default so I don't think it changes anything on your end.oyher than having 2 extra options.

As you said anyone who turned them off when building will do so knowingly.

Having an option to not compile them by default saves some time, if you have to do repeated rebuilds of things where appstream is a dependency. It is not a very significant time saved but I think it is worth providing an option.

The other thing is installed-tests being installed which is being worked around eg. see https://gitlab.alpinelinux.org/alpine/aports/-/blob/4aa55390052999d340a4670a33538f1583f17dcb/community/appstream/APKBUILD#L69

Issue #681, opened 2024-11-11
Files installed through as_utils_install_metadata_file() have executable bit set


mcrha:
When a file is installed through `as_utils_install_metadata_file()`, it has set the executable bit in the destination. See:
```
$ ls -l /var/cache/swcatalog/xml/
-rwxr-xr-x. 1 root root 5719 Nov 11 15:43 rpmfusion-nonfree-steam.xml
```

This is with appstream-1.0.2-4.fc41.x86_64

mcrha:
Hmm, I'm pretty sure the source file does not have set the executable bit, while the destination does have it set (even when it does not exist yet). Your change suggests the problem can be somewhere else, somewhere deeper, because you do not fiddle with the file attributes at all, furthermore you let glib to "inherit the file attributes from the source file", which makes it confusing, like if the problem was in the glib itself.

mcrha:
I'm reopening this, the problem is still present.

mcrha:
See https://github.com/ximion/appstream/pull/694 for my finding.

PR #694, opened 2025-01-16
utils: Remove leftover g_chmod()

mcrha:
The file has set the attributes after commit c6903ab9db7dc702fdff1e4964fab49a3d0de91c, thus this is not needed.

If it was meant to set the executable flag on the parent directory, then it is not needed either, at least according to my tests. The `path_dest` points to the installed file here, not to the parent directory.

Closes https://github.com/ximion/appstream/issues/681


ximion:
I gladly not have this chmod call :-)
Thank you for looking into this!

PR #695, opened 2025-01-16
Various fixes related to Appstream data installation and ZSTD support

mcrha:
There is already the ZSTD compression support built in, but it was not extended into the appstream data installation, which is the place the code will reach it first, because the ZSTD can be used in the repositories.

See the respective commits for the various fixes done along the way.

### DIFF
--- a/src/as-metadata.c
+++ b/src/as-metadata.c
@@ -88,9 +88,13 @@ as_metadata_file_guess_style (const gchar *filename)
 		return AS_FORMAT_STYLE_CATALOG;
 	if (g_str_has_suffix (filename, ".yml.gz"))
 		return AS_FORMAT_STYLE_CATALOG;
+	if (g_str_has_suffix (filename, ".yml.zst"))
+		return AS_FORMAT_STYLE_CATALOG;
 	if (g_str_has_suffix (filename, ".yaml"))
 		return AS_FORMAT_STYLE_CATALOG;
 	if (g_str_has_suffix (filename, ".yaml.gz"))
+		return AS_FORMAT_STYLE_CATALOG;
+	if (g_str_has_suffix (filename, ".yaml.zst"))
 		return AS_FORMAT_STYLE_CATALOG;
 	if (g_str_has_suffix (filename, ".appdata.xml"))
 		return AS_FORMAT_STYLE_METAINFO;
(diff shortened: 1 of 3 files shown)


src/as-metadata.c
```diff
@@ -100,8 +106,6 @@ as_metadata_file_guess_style (const gchar *filename)
 		return AS_FORMAT_STYLE_METAINFO;
 	if (g_str_has_suffix (filename, ".metainfo.xml.in.in"))
 		return AS_FORMAT_STYLE_METAINFO;
-	if (g_str_has_suffix (filename, ".xml"))
```
ximion, line 103: Moving this looks wrong - if a pure match on `.xml` is executed first, the code that's checking if the suffix is e.g. `.metainfo.xml` will never be reached.

ximion (changes requested):
Sorry for the slow review! I found one issue, can you please fix that? Then this PR should be ready to merge :-)

src/as-metadata.c
mcrha: Ouch, you are right. I wanted to have them together, not split (one at the top, another at the bottom) and I did not realize this. I moved the one from the top to the bottom instead.

mcrha: The set of the `if`-s is also kinda funny and slightly harder to read. Maybe this would be prettier? 
```c
	if (g_str_has_suffix (filename, ".xml") ||
	    g_str_has_suffix (filename, ".xml.gz") ||
	    g_str_has_suffix (filename, ".xml.zst"))
		return AS_FORMAT_STYLE_CATALOG;
```

aka make them clear groups. I can extend the change, if you want/agree.

ximion: I actually find the more compact larger if-block you propose harder to read - but only slightly ^^
(I tend to avoid extremely long if-conditions if there is a better option, like bailing out early)

So, I have a slight preference for the current style, but that's it, I also see advantages in the compact version.
tl;dr: I wouldn't object the change, it's just a style choice and either is readable.

ximion (approved):

src/as-metadata.c
ximion: Thanks :-) I had a slight preference for the existing style because it's used everywhere else for these functions as well, and I value consistency quite a lot. But yeah, minor thing :-)

ximion:
Thank you for the patch! :-)

PR #701, opened 2025-03-09
qt: don't dereference error if there is no error

hsitter:
blind fix for https://bugs.kde.org/show_bug.cgi?id=500691

not sure it even builds :)


ximion:
Is there any way I can reproduce this issue? I haven't managed to do that yet and there is no possible way for the function to return an empty error while also returning FALSE.
I reviewed that code, and there is no code path to that that I found. The patch to the Qt bindings papers over a potential issue elsewhere, and I would much rather find the root cause of the problem.

Also, knowing the versions of appstream and libxmlb that the reporter used would be helpful.

PR #706, opened 2025-03-31
qt: Expose markup conversion utils

ismailof:
Expose `as_markup_covert()` function and `AsMarkUpKind` enum, so Qt clients can use the description fields in other suitable formats, such as markdown or simple text

### DIFF
--- a/qt/utils.cpp
+++ b/qt/utils.cpp
@@ -22,6 +22,8 @@
 #include "appstream.h"
 #include "chelpers.h"
 
+#include <QDebug>
+
 QString AppStream::Utils::currentAppStreamVersion()
 {
     return QString::fromUtf8(as_version_string());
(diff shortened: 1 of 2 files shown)


ximion:
This looks good, but is there any better way to send the error back to the caller, rather than to dump it on the console as a sideeffect?

ismailof:
> This looks good, but is there any better way to send the error back to the caller, rather than to dump it on the console as a sideeffect?

I've been trying to find some good way to do it, while keeping the API still convenient, leaning into an optional parameter, maybe a `std::error_code`.

But trying to understand which kind of errors can be produced, it seems the `error` parameter is not used at all in the original function: https://github.com/ximion/appstream/blob/cac6772b6bec50ca3660eed8b6cb6864ee5b6b9d/src/as-utils.c#L216, so maybe it is not even necessary to track it here?

ximion:
The error parameter exists so we can add an error in future in case there are parsing errors that we want to handle, and can add them without breaking API. Clients now would likely simply show the `error->message` if any error was set, and not look for an error code to be on the safe side.

The most straightforward way to convey this would be to emit a C++ exception, but introducing an exception into a library that otherwise doesn't use them would be a bad idea and quite surprising for users of the library.

ximion:
That you for the patch! I will have a look if it's possible to use `std::expected` here for error handling, or whether that is a too steep requirement (GCC 12 supports it, which is the baseline for AppStream, so it might be fine).

ismailof:
> I will have a look if it's possible to use `std::expected` here for error handling, or whether that is a too steep requirement (GCC 12 supports it, which is the baseline for AppStream, so it might be fine).

Thanks for merging! I struggled to find a good way to expose the error handling (idiomatic and convenient) but `std::expected`, which being c++23 seemed a high requirement. IIRC in Discover and KDE Frameworks we now generally depend on c++20

ximion:
Yes, while I think `std::expected` is generally worth it, for just one function it is a very steep requirement increase. So I ultimately went with a different solution, that only needs C++20 :-)

PR #708, opened 2025-04-11
docs: Make docs load in 'devhelp' app

sidt4:
docs: Make docs load in 'devhelp' app

`devhelp` app expects `.devhelp` index files in one of the following locations:

```
/usr/share/doc/appstream/appstream.devhelp2
/usr/share/doc/appstream/appstream.devhelp2.gz
/usr/share/doc/appstream/appstream.devhelp
/usr/share/doc/appstream/appstream.devhelp.gz

/usr/share/gtk-doc/html/appstream/appstream.devhelp2
/usr/share/gtk-doc/html/appstream/appstream.devhelp2.gz
/usr/share/gtk-doc/html/appstream/appstream.devhelp
/usr/share/gtk-doc/html/appstream/appstream.devhelp.gz
```

but we have the index file in the following locations, which doesn't work.

```
/usr/share/doc/appstream/html/api/api.devhelp2
/usr/share/gtk-doc/html/appstream/api.devhelp2 (symlink to above)
```

Refer following comment in https://github.com/GNOME/devhelp/commit/72f8e5f60.

```
/* The name of the directory the index file is in and the name of
the index file (minus the extension) must match. */
```

### DIFF
--- a/docs/api/compose/meson.build
+++ b/docs/api/compose/meson.build
@@ -15,7 +15,7 @@ custom_target('doc-api-appstream-compose',
       ascdoc_toml,
       asc_gir,
     ],
-    output: 'compose-api',
+    output: 'appstream-compose',
     command: [
       gidocgen_exe,
       'generate',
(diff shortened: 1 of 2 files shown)


ximion:
I am pretty sure this will break *all* links, so we can't do that. But the issue is real, I'll have a look, maybe we can rename things post-install or do something sensible with symlinks (I'd also have to play around with it).

sidt4:
> I am pretty sure this will break all links, so we can't do that. 

You mean URLs in https://www.freedesktop.org/ (which I guess can be updated, no?).

ximion:
This should probably be addressed in gi-docgen, but I added a small wrapper script that allows setting the namespace directly and fixes the devhelp file misnaming.
Hopefully that solves the issue :-)

sidt4:
Works. Thanks!

PR #714, opened 2025-06-08
compose: Accept .ttc font files

joebonrichie:
Most common with CJK fonts

### DIFF
--- a/compose/asc-utils-fonts.c
+++ b/compose/asc-utils-fonts.c
@@ -649,7 +649,9 @@ asc_process_fonts (AscResult *cres,
 		const gchar *fname = g_ptr_array_index (contents, i);
 		if (!g_str_has_prefix (fname, fonts_dir))
 			continue;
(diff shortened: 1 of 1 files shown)


ximion:
> Screenshots and icons seems to get generated correctly with the path

Nice, that's what I needed to know! Normally a test for this would be nice, but this is hard to test anyway and we can't ship a ton of random fonts and compare the result, so, this patch is fine for now.

Thank you!

joebonrichie:
We may need to adjust the fontconfig APIs to handle .ttc files correctly elsewhere, e.g. to iterate to the index of all fonts contained within the .ttc,

here is a bash one-liner for it
`INDEX=`fc-query -f '%{index}\n' my-font-file.ttc | sort -nu | tail -n1`; for i in $(seq 1 "$INDEX"); do fc-query --index="$i" --format='%{fullname[0]}\n' my-font-file.ttc; done`

PR #716, opened 2025-06-09
compose: Redetermine sample icon text if sample is too long

joebonrichie:
Some fonts are set with a sample text that is way to long to be legible, redetermine the icon text if length > 3.


ximion:
Looks like font samples coming from the font data itself were not sanitized. I fixed that, but it's still odd that this data was actually used in the first place...

Can you check if the patch already helps and how the result looks like?

Issue #717, opened 2025-06-10
compose: Text may overlap canvas background shape for auto-generated font icons at most sizes


joebonrichie:
Since 9b1c6466e591d5da0ed23e10ba758fb726fd87f1

There are alignment issues with some fonts where the text overlaps the edges of the canvas background shape, this is particularly common with the hexagon shape. Note, this affects most sizes but occasionally the 128x128@2 size has no overlap.

Examples:
iosevka : 
128x128: https://appstream.getsol.us/media/pool/io/github/be5invis.iosevka/1405e39cb2559fedefc6fa263822c5f9/icons/128x128/font-iosevka-ttf_iosevka-regular.png
128x128@2 (no overlap): https://appstream.getsol.us/media/pool/io/github/be5invis.iosevka/1405e39cb2559fedefc6fa263822c5f9/icons/128x128@2/font-iosevka-ttf_iosevka-regular.png
64x64: https://appstream.getsol.us/media/pool/io/github/be5invis.iosevka/1405e39cb2559fedefc6fa263822c5f9/icons/64x64/font-iosevka-ttf_iosevka-regular.png

cantarell:
128x128: https://appstream.getsol.us/media/pool/org/gnome/cantarell/4c38d602fb128d0a2feca1f75d57cee7/icons/128x128/cantarell-fonts_cantarell-regular.png
128x128@2: https://appstream.getsol.us/media/pool/org/gnome/cantarell/4c38d602fb128d0a2feca1f75d57cee7/icons/128x128@2/cantarell-fonts_cantarell-regular.png
64x64: https://appstream.getsol.us/media/pool/org/gnome/cantarell/4c38d602fb128d0a2feca1f75d57cee7/icons/64x64/cantarell-fonts_cantarell-regular.png

PR #718, opened 2025-06-10
compose: Trim substring when truncating font sample text

joebonrichie:
This is to avoid alignment issues when presenting text, e.g. font icons

Before:
install-_overpass-regular

After:
install-_overpass-regular

Part of #717


ximion (changes requested):

ximion:
Ah, sorry, I didn't see you made the requested changes in the latest commit changes! This looks good now, besides some code formatting issues which I can fix in a subsequent commit.

Thank you for the patch!

PR #721, opened 2025-07-30
meson: don't override appstreamcli when cross-building

elmarco:
Using the target build require an exe-wrapper.


ximion:
Makes sense, thank you!

PR #725, opened 2025-08-12
Minor Grammatical Fix/Improvement

chrisdenman:
Just a slight grammatical tweak. Maybe 'consists of' would suit you better.


ximion:
You need to make this change in the C code, any change to the translation files will be overridden either during the next template regeneration or, in this case, by Weblate.

PR #761, opened 2026-06-28
FreeBSD CI: Switch from Cirrus to GitHub Action

arrowd:


.github/workflows/build-test.yml
```diff
@@ -130,3 +130,54 @@ jobs:
...
+          echo "::group::Dependencies installation"
+          sudo -E pkg install -y \
+              gettext-tools glib gobject-introspection gperf gtk-doc \
+              itstool libxmlb libfyaml meson pkgconf qt6-tools \
+              snowballstemmer libblake3 textproc/py-gi-docgen vala
+          echo "::endgroup::"
+
+          # Workaround the 'warning: failed to load external entity "http://docbook.sourceforge.net/release/xsl-ns/current/manpages/docbook.xsl"'
+          # error when calling xsltproc
+          sed -i.bak -e 's|/xsl-ns/|/xsl/|g' docs/meson.build
```
ximion: Does that actually generate usable manpages? If so, instead of this hack, we can maybe find a better solution in the `meson.build` file itself...

arrowd: Yes. It fails because on FreeBSD the catalog file `/usr/local/share/xsl/docbook/catalog.xml` contains the following line:
```
<rewriteURI uriStartString="http://docbook.sourceforge.net/release/xsl/current/" rewritePrefix="./"/>
```
Note the `/xsl-ns/` vs `/xsl/` difference. I did some research, but failed to figure out who's wrong here, so resorted to patching.

ximion: I just checked with the spec, FreeBSD is wrong. We do require the namespaced version.
A reason for this could actually be that you are hitting Docbook 4, not Docbook 5. In any case, this hack is wrong and most likely breaks things.

.github/workflows/build-test.yml
```diff
@@ -130,3 +130,54 @@ jobs:
...
+      - name: Run
+        run: |
+          # Use latest package set
+          sudo mkdir -p /usr/local/etc/pkg/repos/
+          sudo cp /etc/pkg/FreeBSD.conf /usr/local/etc/pkg/repos/FreeBSD.conf
+          sudo sed -i.bak -e 's|/quarterly|/latest|' /usr/local/etc/pkg/repos/FreeBSD.conf
+
+          echo "::group::Dependencies installation"
+          sudo -E pkg install -y \
+              gettext-tools glib gobject-introspection gperf gtk-doc \
```
ximion:
```suggestion
              gettext-tools glib gobject-introspection gperf textproc/docbook-xsl-ns \
```

I found https://www.freshports.org/textproc/docbook-xsl-ns , that should work. AppStream also doesn't use gtk-doc anymore for quite a while, and gi-docgen is actually installed properly.

ximion (changes requested):

.github/workflows/build-test.yml
ximion: Adding `textproc/docbook-xsl-ns` and dropping that hack should most likely fix it, it's work a try! :-) With this hack, you will most certainly get broken manpages.

ximion:
Can you push again? For some reason, it didn't actually run the modified GitHub Action (which may actually be a security thing by GitHub, in which case I should maybe just merge this and see what happens...)

.github/workflows/build-test.yml
arrowd: Thanks for teaching me something new. I didn't know about that xsl/xsl-ns distinction before. Your suggested change does work indeed.

arrowd:
Yeah, the build doesn't run in this PR for some reason, but it runs in my fork. For some reason one test fails now, I'll take a look later.

ximion:
Okay, maybe I should wait with merging this then until tests pass. The patch itself looks fine, and I am happy to have everything in Actions now. The `LDFLAGS` and `CFLAGS` stuff set externally looks kinda strange to me, is this a FreeBSD thing that's necessary? Nothing in it looks like it would result in any issue at all though :-)

ximion:
That test failure is related to libfyaml changes, not to this change, so I merged this commit (and we'll have to address that libfyaml change separately).

PR #772, opened 2026-07-30
qt/tests: add appstream_dep to the test executables

dfaure:
With only link_with: appstreamqt_lib, meson adds no link-time search path for the build tree's libappstream, so the linker resolves the indirect dependency from the install prefix instead. That fails whenever the installed copy is older than the API AppStreamQt uses.

I got `/usr/bin/ld.bfd: qt/libAppStreamQt.so.1.1.6: undefined reference to 'as_issue_get_json_url'`

### DIFF
--- a/qt/tests/meson.build
+++ b/qt/tests/meson.build
@@ -18,7 +18,7 @@ asqt_test_pool_src += qt.preprocess (moc_sources: asqt_test_pool_src)
 
 as_test_qt_pool_exe = executable('as-test_qt-pool',
     [asqt_test_pool_src],
-    dependencies: [qt_test_dep],
+    dependencies: [qt_test_dep, appstream_dep],
     include_directories: [include_directories('..')],
     link_with: [appstreamqt_lib],
     cpp_args: asqt_cpp_args,
@@ -36,7 +36,7 @@ asqt_test_misc_src += qt.preprocess (moc_sources: asqt_test_misc_src)
 
 as_test_qt_misc_exe = executable('as-test_qt-misc',
     [asqt_test_misc_src],
-    dependencies: [qt_test_dep],
+    dependencies: [qt_test_dep, appstream_dep],
     include_directories: [include_directories('..')],
     link_with: [appstreamqt_lib],
     cpp_args: asqt_cpp_args,


ximion:
I don't particularly like this, because the binaries must not use any of `libappstream`'s symbols.

I am building the Qt bindings just fine even though an older version of libappstream is installed on the system that does not have the same symbols. So, do you know why your Meson tries to link with a system copy of libas instead of the just-built one? I tested a few permutations, and Meson/Ninja always invoke this with the correct version on my system (Meson 1.11.1).

jpetso:
For the record, I also encountered this on a clean rebuild of appstream on my Arch system. (Also on an incremental build first, before I nuked all the local build products.) This was the `tail` end of the build file:

```
[258/261] Compiling C++ object qt/tests/as-test_qt-pool.p/asqt-pool-test.cpp.o
[259/261] Compiling C++ object qt/tests/as-test_qt-misc.p/asqt-misc-test.cpp.o
[260/261] Linking target qt/tests/as-test_qt-pool
FAILED: [code=1] qt/tests/as-test_qt-pool 
c++  -o qt/tests/as-test_qt-pool qt/tests/as-test_qt-pool.p/asqt-pool-test.cpp.o -Wl,--as-needed -Wl,--no-undefined '-Wl,-rpath,$ORIGIN/..:$ORIGIN/../../src' -Wl,--start-group qt/libAppStreamQt.so.1.1.6 /usr/lib/libQt6Core.so /usr/lib/libQt6Test.so -Wl,--end-group
/usr/bin/ld: qt/libAppStreamQt.so.1.1.6: undefined reference to `as_issue_get_json_url'
collect2: error: ld returned 1 exit status
[261/261] Linking target qt/tests/as-test_qt-misc
FAILED: [code=1] qt/tests/as-test_qt-misc 
c++  -o qt/tests/as-test_qt-misc qt/tests/as-test_qt-misc.p/asqt-misc-test.cpp.o -Wl,--as-needed -Wl,--no-undefined '-Wl,-rpath,$ORIGIN/..:$ORIGIN/../../src' -Wl,--start-group qt/libAppStreamQt.so.1.1.6 /usr/lib/libQt6Core.so /usr/lib/libQt6Test.so -Wl,--end-group
/usr/bin/ld: qt/libAppStreamQt.so.1.1.6: undefined reference to `as_issue_get_json_url'
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.

# exit code was: 256
```

Whereas with `appstream_dep` added manually, the libraries are explicitly expanded when looking at the generated `build.ninja`:

```make
build qt/tests/as-test_qt-pool: cpp_LINKER qt/tests/as-test_qt-pool.p/meson-generated_.._.._.._src_as-enum-types.c.o qt/tests/as-test_qt-pool.p/asqt-pool-test.cpp.o | /usr/lib/libQt6Core.so /usr/lib/libQt6Test.so /usr/lib/libglib-2.0.so /usr/lib/libgobject-2.0.so qt/libAppStreamQt.so.1.1.6.p/libAppStreamQt.so.1.1.6.symbols src/libappstream.so.1.1.6.p/libappstream.so.1.1.6.symbols
 LINK_ARGS = -Wl,--as-needed -Wl,--no-undefined '-Wl,-rpath,$$ORIGIN/..:$$ORIGIN/../../src' -Wl,--start-group qt/libAppStreamQt.so.1.1.6 src/libappstream.so.1.1.6 /usr/lib/libglib-2.0.so /usr/lib/libgobject-2.0.so /usr/lib/libQt6Core.so /usr/lib/libQt6Test.so -Wl,--end-group
```

However, I'm not enough of a linker expert to know why the test executables in `qt/tests/` would not use the shared libraries from rpath `$ORIGIN/../../src` unless explicitly passed on the command line.

Edit: versions on Arch as of 2026-08-03:
* Meson 1.11.2
* c++ (GCC) 16.1.1 20260728
* GNU ld (GNU Binutils) 2.47
* system-wide /usr/lib/libappstream.so.1.1.5

ximion:
The linker should prefer the local version... And given that CI succeeds, it clearly knows where to find the library. And even with your command-line, the local version should be preferred. This is very, very strange. Linker args also look sane to me.

Just for completeness, you don't have any `LD_*` env vars in your environment, right?
Also, this is 100% reproducible for you on Arch? (I tried many times on Debian and I never saw an issue even once) If Arch fails reliably, we can debug this fairly nicely, I will try later in a chroot.

With this kind of stuff, I definitely want to learn the root cause and solve it - no band-aid on linker behavior ever stays unpunished long-term :wink: .

ximion:
This is a Meson bug. See https://github.com/mesonbuild/meson/issues/16087
I will implement a workaround.

Thank you for reporting the issue and coming up with a patch! :-)

dfaure:
I see now that the meson bug report is asking if LD_LIBRARY_PATH was set. The answer is NO. I want to rely on DT_RUNPATH instead (one reason is: so that I can run tests from the builddir without them picking up older installed versions of the libs).
To be extra sure, I just reproduced the bug, and it happens without LD_LIBRARY_PATH set.
```
$ cd <srcdir>; git switch -c old a3183bdbfb1ad671f2e540c47e8972368f377d44~
$ cd <builddir>; ninja install
$ cd <srcdir>; git switch main
$ cd <builddir>; ninja
FAILED: [code=1] qt/tests/as-test_qt-pool
/usr/bin/ld.bfd: qt/libAppStreamQt.so.1.1.6: undefined reference to `as_issue_get_json_url'

$ grep -A2 as-test_qt-pool: build.ninja
build qt/tests/as-test_qt-pool: cpp_LINKER qt/tests/as-test_qt-pool.p/asqt-pool-test.cpp.o | /d/qt/6/inst-optimized/lib/libQt6Core.so /d/qt/6/inst-optimized/lib/libQt6Test.so qt/libAppStreamQt.so.1.1.6.p/libAppStreamQt.so.1.1.6.symbols
 LINK_ARGS = -Wl,--as-needed -Wl,--no-undefined '-Wl,-rpath,$$ORIGIN/..:$$ORIGIN/../../src:/d/qt/6/inst-optimized/lib:/d/kde/inst/kde_frameworks6/lib64' -Wl,--start-group qt/libAppStreamQt.so.1.1.6 /d/qt/6/inst-optimized/lib/libQt6Core.so /d/qt/6/inst-optimized/lib/libQt6Test.so -Wl,--end-group
```

I don't understand what `$$ORIGIN/..:$$ORIGIN/../../src` is about.
Shouldn't it be `$$ORIGIN/../src`, in order to go from `<builddir>/tests` (where `as-test_pool` lives) to `<builddir>/src` (where `libappstream.so` lives) ?

PR #785, opened 2026-09-02
Do not hard-code libstemmer include directories.

### LINKED ISSUE #569: Don't hardcode /usr/include in meson.build
If AppStream is build with stemmer enabled, it forcibly looks in `/usr/include`:

https://github.com/ximion/appstream/blob/1217913bc0276ad5a0bf02b8ab93744c373fc4f4/meson.build#L188

I may have installed stemmer anywhere, and in cross-compiled environments hardcoding this path is absolutely the wrong thing to do as it is the _host_ includes not the _target_ includes.

apteryks:
Instead, assume `libstemmer.h` is already available on the include path
of the compiler, else point a user to a new `libstemmer_include_dir` option.

Fixes: #569

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -228,14 +228,15 @@ endif
 
 stemmer_inc_dirs = include_directories()
 if get_option('stemming')
+    libstemmer_include_dir = get_option ('libstemmer_include_dir')
+    if libstemmer_include_dir != ''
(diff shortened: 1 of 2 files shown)


ximion:
I solved this slightly differently, but thank you for the patch! Try Git master, that should work to resolve the issue.

apteryks:
I just tried it, it works too! Thank you for the fix.

PR #786, opened 2026-09-02
Add support for elogind.

apteryks:
Elogind provides the systemd/sd-device.h and systemd/sd-hwdb.h
headers, can be used as a drop-in replacement of systemd.

* meson.build ([systemd]): Add elogind alternative name for dependency check.
* meson_options.txt [systemd]: Mention elogind in description.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -202,7 +202,7 @@ else
     zstd_dep = dependency('', required: false)
 endif
 if get_option('systemd')
-    libsystemd_dep = dependency('libsystemd', required: get_option('systemd'))
+    libsystemd_dep = dependency('libsystemd', 'libelogind', required: get_option('systemd'))
 else
     libsystemd_dep = dependency('', required: false)
 endif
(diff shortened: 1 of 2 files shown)


ximion:
AppStream does not use any of the login API. It uses libsystemd to interface with udev and read the hwdb to identify devices on the system.

The elogind docs say:
> Basically all symbols are included. But any API calls that require to call systemd, or need internal knowledge of systemd, are simple stubs. They are there to provide ABI compatibility, but will not work.

That lets me believe that when building with it, we will just end up with completely broken functionality. Under these circumstances, isn't it better to just build without systemd when it isn't available? At least that will provide an obvious error message, while reading stub data will result in permission warnings or odd failure messages that may be harder to reason about.

apteryks:
elogind implements/imports the bits of systemd needed for modern desktops like GNOME, which is a larger subset than just logind; I've reviewed the source of elogind; its hwdb support makes use of udev (which it links to) for example. These are not stubs.

ximion:
Okay, that wasn't clear to me. In that case, this patch is fine, thanks!

PR #788, opened 2026-09-02
system-info: Avoid overflow in physical memory total on 32-bit systems.

apteryks:
* src/as-system-info.c (as_get_physical_memory_total): Add a guint64
cast to ensure the multiplication is done using 64-bit arithmetic.


ximion:
Thank you! :-)

